PR #4, opened 2018-09-14
Rename localized conf file to tony-final.xml

hungj:
Verified confs on CLI still override those in tony.xml. Also verified confs are propagated to AM/TaskExecutors.


erwa:
Thanks for making this change. Looks good to me.

PR #13, opened 2018-09-19
A Notebook submitter to run Jupyter Notebook inside cluster

oliverhu:


tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -670,6 +675,7 @@ private int doPreprocessingJob() throws Exception {
     }
 
     extraEnv.put(Constants.PREPROCESSING_JOB, "true");
+    extraEnv.put("HOME", System.getProperty("user.dir"));
```
erwa: What is the purpose of setting `HOME`?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -729,7 +735,17 @@ public void reset() {
 
     @Override
     public Set<TaskUrl> getTaskUrls() {
-      if (session != null && session.allTasksScheduled()) {
+      LOG.info("Client requesting TaskUrls!");
+      if (singleNode && proxyUrl != null) {
+        return new HashSet<TaskUrl>(){{
```
erwa: I suggest avoiding double-brace initialization. It's generally not a good Java practice: https://stackoverflow.com/a/27521360/1128392

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -665,7 +672,12 @@ private void printTaskUrls() throws IOException, YarnException {
     if (amRpcServerInitialized && !hasPrintedTaskUrls) {
       Set<TaskUrl> taskUrls = amRpcClient.getTaskUrls();
       if (!taskUrls.isEmpty()) {
-        new TreeSet<TaskUrl>(taskUrls).forEach(task -> Utils.printTaskUrl(task, LOG));
+        new TreeSet<TaskUrl>(taskUrls).forEach(task -> {
+            if (task.getName().equals(Constants.NOTEBOOK_JOB_NAME)) {
+                notebookUrl = task.getUrl();
```
erwa: This `printTaskUrls()` method shouldn't have side effects like setting a `notebookUrl` field. Please do this in a separate method.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -701,6 +713,18 @@ private void cleanUp() {
     }
   }
 
+  public static TonyClient createClientInstance(String[] args, Configuration conf) throws ParseException {
+    TonyClient client;
+    client = new TonyClient(conf);
+    boolean sanityCheck = client.init(args);
+    client.createYarnClient();
```
erwa: Can `createYarnClient()` be done as part of `init()` or when `run()` is called on the client? Seems a little weird to call this here.

Also, I think `sanityCheck` should be checked before this line.

tony-core/src/test/java/com/linkedin/tony/TestUtils.java
```diff
@@ -40,6 +40,27 @@ public void testUnzipArchive() {
     } catch (IOException e) {
       fail(e.toString());
     }
+  }
 
+  @Test
+  public void testIsArchive() {
```
erwa: Can you also add some negative tests (where `isArchive()` returns `false`)?

tony-core/src/test/java/com/linkedin/tony/TestUtils.java
```diff
@@ -40,6 +40,27 @@ public void testUnzipArchive() {
...
+    File file1 = new File(classLoader.getResource("test.zip").getFile());
+    File file2 = new File(classLoader.getResource("test.tar").getFile());
+    File file3 = new File(classLoader.getResource("test.tar.gz").getFile());
+    assertTrue(Utils.isArchive(file1.getAbsolutePath()));
+    assertTrue(Utils.isArchive(file2.getAbsolutePath()));
+    assertTrue(Utils.isArchive(file3.getAbsolutePath()));
+  }
+
+  @Test
+  public void testRenameFile() throws IOException {
```
erwa: If this test is run multiple times, will it fail on subsequent attempts, since the `file_move.txt` resource file will have been moved? I think it's better to create a temp file and move that instead.

tony-core/src/test/java/com/linkedin/tony/TestUtils.java
```diff
@@ -40,6 +40,27 @@ public void testUnzipArchive() {
...
       fail(e.toString());
     }
+  }
 
+  @Test
+  public void testIsArchive() {
+    ClassLoader classLoader = getClass().getClassLoader();
+    File file1 = new File(classLoader.getResource("test.zip").getFile());
+    File file2 = new File(classLoader.getResource("test.tar").getFile());
+    File file3 = new File(classLoader.getResource("test.tar.gz").getFile());
```
erwa: Can the test generate these .zip, .tar, and .tar.gz files instead of checking them in? (In general, I think we should reduce the number of binary artifacts checked in to our repo.)

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: Because YARN sets HOME to `/` and that is not the right home directory for our job. We don't see this problem since we normally don't access HOME. However.. Jupyter Notebook wants to write some cache files to $HOME and that results to permission error.

tony-proxy/src/main/java/com/linkedin/tonyproxy/ProxyServer.java
```diff
@@ -0,0 +1,110 @@
...
+    this.host = host;
+    this.remotePort = remotePort;
+    this.localPort = localPort;
+  }
+  public void start() throws IOException {
+    LOG.info("Starting proxy for " + host + ":" + remotePort
+             + " on port " + localPort);
+    ServerSocket server = new ServerSocket(localPort);
+    while (true) {
+      new Proxy(server.accept(), host, remotePort).start();
```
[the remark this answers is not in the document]
oliverhu: This blocks till a new request comes in.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
oliverhu: yah.. this was a hack, will clean this up.

tony-core/src/test/java/com/linkedin/tony/TestUtils.java
oliverhu: We don't have the utils to generate archive files.. also, we need different logic/dependencies to zip/tar/tar.gz files, that is overkill for this.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
erwa: Got it. Can you add a code comment about this?

tony-core/src/test/java/com/linkedin/tony/TestUtils.java
erwa: Fair enough

tony-cli/src/main/java/com/linkedin/tony/cli/NotebookSubmitter.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+    if (cachedLibPath == null) {
+      System.exit(-1);
+    }
+
+    String[] updatedArgs = Arrays.copyOf(args, args.length + 4);
+    updatedArgs[args.length] = "--hdfs_classpath";
+    updatedArgs[args.length + 1] = cachedLibPath.toString();
+    updatedArgs[args.length + 2] = "--conf";
+    updatedArgs[args.length + 3] = TonyConfigurationKeys.APPLICATION_TIMEOUT + "=" + String.valueOf(24 * 3600 *1000);
```
hungj: nit: 24 * 60 * 60 * 1000

tony-cli/src/main/java/com/linkedin/tony/cli/NotebookSubmitter.java
```diff
@@ -0,0 +1,98 @@
...
+    });
+    clientThread.start();
+    while (clientThread.isAlive()) {
+      if (client.notebookUrl != null) {
+        String[] hostPort = client.notebookUrl.split(":");
+        ServerSocket localSocket = new ServerSocket(0);
+        int localPort = localSocket.getLocalPort();
+        localSocket.close();
+        ProxyServer server = new ProxyServer(hostPort[0], Integer.parseInt(hostPort[1]), localPort);
+        LOG.info("Please run [ssh -L 8080:localhost:" + String.valueOf(localPort) + " name_of_this_host] in your laptop and open [localhost:8080] in your browser to visit Notebook");
```
hungj: should we make 8080 configurable? maybe they have stuff running on this port locally already

tony-core/src/main/java/com/linkedin/tony/Constants.java
```diff
@@ -48,6 +46,8 @@
 
   public static final String WORKER_JOB_NAME = "worker";
   public static final String PS_JOB_NAME = "ps";
+  public static final String NOTEBOOK_JOB_NAME = "Notebook";
```
hungj: Can we lowercase Notebook and Driver for consistency?

tony-proxy/src/main/java/com/linkedin/tonyproxy/ProxyServer.java
```diff
@@ -0,0 +1,110 @@
...
+    private final int serverPort;
+    Proxy(Socket client, String host, int port) {
+      this.clientSocket = client;
+      this.serverHost = host;
+      this.serverPort = port;
+    }
+
+    @Override
+    public void run() {
+      try {
```
hungj: can we use try-with-resources for inFromClient/outToClient/inFromServer/outToServer?

tony-core/src/main/java/com/linkedin/tony/Constants.java
```diff
@@ -48,6 +46,8 @@
...
+  public static final String DRIVER_JOB_NAME = "Driver";
```
hungj: Not sure I understand what this is used for?

tony-cli/src/main/java/com/linkedin/tony/cli/NotebookSubmitter.java
oliverhu: 8080 is only used in logging.. maybe I can explain a bit more here like: try 8080 otherwise use another port

tony-proxy/src/main/java/com/linkedin/tonyproxy/ProxyServer.java
oliverhu: 👌

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -456,7 +458,12 @@ public ContainerLaunchContext createAMContainerSpec(ApplicationId appId, String
   private void zipArchive() throws IOException {
     FileOutputStream fos = new FileOutputStream(ARCHIVE_PATH);
     ZipOutputStream zos = new ZipOutputStream(fos);
-    addDirToZip(zos, srcDir);
+    // Accept archive file as srcDir.
+    if (!Utils.isArchive(srcDir)) {
+      addDirToZip(zos, srcDir);
+    } else {
+      Utils.renameFile(srcDir, ARCHIVE_PATH);
```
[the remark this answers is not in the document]
oliverhu: the logic is if it is an Archive file, we rename it..

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: Good to know, thanks!

tony-core/src/test/java/com/linkedin/tony/TestUtils.java
oliverhu: good catch!

tony-core/src/main/java/com/linkedin/tony/Constants.java
oliverhu: This is used for sending the address:port of notebook back to TonyClient

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
oliverhu: yah.. this was also some tmp workaround to get the client, gonna clean this up as well.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -637,17 +646,16 @@ private void addClientToAMTokenToUGI(ApplicationReport report) throws IOExceptio
     }
   }
 
-  private void printTaskUrls() throws IOException, YarnException {
-    if (amRpcServerInitialized && !hasPrintedTaskUrls) {
-      Set<TaskUrl> taskUrls = amRpcClient.getTaskUrls();
+  private void getApplicationTaskUrls() throws IOException, YarnException {
```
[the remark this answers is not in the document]
oliverhu: totally right..

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -105,13 +105,18 @@
...
+
   private static final String ARCHIVE_PATH = "tf_archive.zip";
   private Configuration tonyConf;
   private final long clientStartTime = System.currentTimeMillis();
   private Path appResourcesPath;
   private int hbInterval;
   private int maxHbMisses;
 
+  // For access from CLI.
+  public Set<TaskUrl> taskUrls = new HashSet<>();
```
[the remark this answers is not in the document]
oliverhu: makes sense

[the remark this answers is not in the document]
oliverhu: this doesn't look performant tho? You'd have to use `ImmutableSet.copyOf(xxx)` which copies everything ?

oliverhu:
validated functionality again after all the updates and it works, merging it!

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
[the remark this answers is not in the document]
erwa: True, it does copy everything, but I doubt this will ever be the performance bottleneck.

tony-proxy/src/main/java/com/linkedin/tonyproxy/ProxyServer.java
```diff
@@ -52,57 +52,40 @@ public void start() throws IOException {
 
     @Override
     public void run() {
-      try {
+      try (Socket server = new Socket(serverHost, serverPort);
+          final InputStream inFromClient = clientSocket.getInputStream();
+          final OutputStream outToClient = clientSocket.getOutputStream();
+          final InputStream inFromServer = server.getInputStream();
+          final OutputStream outToServer = server.getOutputStream();
```
erwa: Code looks much cleaner with the try-with-resources construct!

oliverhu: Yep

PR #15, opened 2018-09-20
Add Azkaban module with TensorFlowJob

hungj:


tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
```diff
@@ -0,0 +1,195 @@
...
+        props.put(HADOOP_OPTS, String.format("%s %s", hadoopOps, hadoopGlobalOps));
+      } else {
+        props.put(HADOOP_OPTS, hadoopGlobalOps);
+      }
+    }
+  }
+
+  @Override
+  public String getJavaClass() {
+    return "com.linkedin.tony.TonyClient";
```
oliverhu: Switch to ClusterSubmitter

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
```diff
@@ -0,0 +1,195 @@
...
+  protected String getMainArguments() {
+    String args = super.getMainArguments();
+    info("All job props: " + getJobProps());
+    info("All sys props: " + getSysProps());
+    String srcDir = getJobProps().getString(TensorFlowJobArg.SRC_DIR.azPropName, "src");
+    args += " " + TensorFlowJobArg.SRC_DIR.tfParamName + " " + srcDir;
+
+    String hdfsClasspath = getJobProps().getString(TensorFlowJobArg.HDFS_CLASSPATH.azPropName, null);
+    if (hdfsClasspath != null) {
+      args += " " + TensorFlowJobArg.HDFS_CLASSPATH.tfParamName + " " + hdfsClasspath;
```
oliverhu: HDFS_CLASSPATH should be deprecated

build.gradle
```diff
@@ -23,33 +23,43 @@ buildscript {
...
-        "sshd": "org.apache.sshd:sshd-core:1.1.0",
-        "testng": "org.testng:testng:6.4",
-        "text": "org.apache.commons:commons-text:1.4",
-        "zip4j": "net.lingala.zip4j:zip4j:1.3.2"
-    ]
+  azkaban: [
+    "az_core": "com.linkedin.azkaban:az-core:3.58.0",
+    "azkaban_common": "com.linkedin.azkaban:azkaban-common:3.58.0",
+    "azkaban_plugins": "com.linkedin.azkaban:az-hadoop-jobtype-plugin:3.58.0",
+    "azkaban_security": "com.linkedin.azkaban:azkaban-hadoop-security-plugin:3.58.0"
```
erwa: Can we use the full names `az_hadoop_jobtype_plugins` and `azkaban_hadoop_security_plugin` as our variable names? Azkaban has other types of plugins besides jobtype plugins, and also other types of security plugins.

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
```diff
@@ -0,0 +1,195 @@
...
+    List<String> classPath = super.getClassPaths();
+    classPath.add(tonyConfFile.getParent());
+    return classPath;
+  }
+
+  /**
+   * First refresh the extract
+   * once the status is found, log the results and cancel the job if
+   * the status was an error or a timeout
+   */
```
erwa: What is this comment talking about?

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
hungj: I don't think we should deprecate this. In general users should run with hdfs_classpath. I don't think it's a good practice to use the fatjar approach, even if it's easiest.

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
hungj: Same as below...ClusterSubmitter is useful for submitting from gateway quickly, but dependency management is much cleaner if hdfs_classpath is configured by users and uses the TonY thin jar.

build.gradle
hungj: sure, fixed

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
hungj: No clue, removed

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
oliverhu: What is the disadvantage of using a farjar here? it only makes user's life easier. 
- it avoids the jar conflicts problem
- it prevents the problem is a user forgot to run the cachedLib task

hungj: What jar conflict problem are you referring to? Using fatjar makes dependency conflicts worse. e.g if they are running one version of Hadoop which has a dependency with version X, and TonY is built against another version which pulls in that dependency with version Y, there will be a mismatch

PR #18, opened 2018-09-27
Determine TensorFlow job names in user configuration at runtime

hungj:
Currently configurations such as tony.worker.instances are hardcoded to use job names like "worker".

Instead, we want to resolve these configured job names at runtime, so if the user specifies "tony.chief.instances", it will handle this as another TensorFlow job type (i.e. include it in the cluster spec, parse out tony.chief.memory, tony.chief.gpus, etc. if included in the conf). This way arbitrary TensorFlow job names can be requested.


tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -63,7 +63,6 @@
 import org.apache.hadoop.yarn.api.records.UpdatedContainer;
```
oliverhu: Could you update READ.me & Azkaban plugin & provide an example?

PR #29, opened 2018-10-05
Skeleton for Tony History Server

pdtran3k6:
Run history server by `$root/bin/start-history-server.sh`.
Will update logo once we decide.


erwa:
Still need to review a patch, but a quick comment: for the logo, you can use the TonY logo here: https://content.linkedin.com/content/dam/engineering/site-assets/images/blog/posts/2018/09/tony1.jpg

oliverhu:
This PR looks huge.. some more information on how this tony-history-server component is organized in your comment? That would make review easier.

pdtran3k6:
Like we discussed yesterday, this is just a fork of XLearning's history server code with some refactoring, which means it has a lot of bloat/inapplicable classes for TonY right now. Would you rather me keep cleaning it up with following commits until it's more concise or should I just explain all the components as is?
I'll also update the logo.

erwa:
Let's keep cleaning/trimming it down before committing it. Also, can you update the `NOTICE` file to include something like:

```
This product includes/uses XLearning (https://github.com/qihoo360/xlearning)
License:

* Copyright (c) 2017, qihoo
* All rights reserved.
* Redistribution and use in source and binary forms, with or without
* modification, are permitted provided that the following conditions are met:
*
*     * Redistributions of source code must retain the above copyright
*       notice, this list of conditions and the following disclaimer.
*     * Redistributions in binary form must reproduce the above copyright
*       notice, this list of conditions and the following disclaimer in the
*       documentation and/or other materials provided with the distribution.
*     * Neither the name of qihoo nor the names of its contributors may be 
*       used to endorse or promote products derived from this software
*       without specific prior written permission.
*
* THIS SOFTWARE IS PROVIDED BY THE QIHOO AND CONTRIBUTORS "AS IS" AND ANY
* EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
* WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
* DISCLAIMED. IN NO EVENT SHALL THE QIHOO AND CONTRIBUTORS BE LIABLE FOR ANY
* DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

bin/start-history-server.sh
```diff
@@ -0,0 +1,60 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+set -e
+
+export THS_HOME="$(cd "`dirname "$0"`"/..; pwd)" # root of the project
```
erwa: Rather than make things _history server_ specific, let's keep everything as just "tony". We should have a single tony configuration file/home that both clients and the history server can share. This is similar to what MapReduce does. A single `HADOOP_MAPRED_HOME` environment variable and the `mapred-site.xml` file contains both client-side configs and job history server configs.

So I think we should name things like:

* TONY_HOME
* tony-env.sh
* tony-site.xml
* etc.

tony-history-server/src/main/resources/tony-default.xml
```diff
@@ -0,0 +1,192 @@
+<?xml version="1.0"?>
```
erwa: This file, and the `tony-history-server/src/main/resources/log4j.properties` look copied from `tony-core`. Why do you need these here?

tony-history-server/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+  compile deps.hadoop.mapreduce_client_common
+  compile deps.hadoop.mapreduce_client_app
+  testCompile deps.external.avro
+  testCompile deps.external.testng
+
+  // Only needed by Hadoop test classes
+  testRuntime deps.external.junit
+}
+
+task setupHdfsLib(type: Copy) {
```
erwa: This task doesn't seem necessary in the tony-history-server module. Looks like leftover from copying the tony-core build.gradle file over.

pdtran3k6:
> Let's keep cleaning/trimming it down before committing it. Also, can you update the `NOTICE` file to include something like:
> 
> ```
> This product includes/uses XLearning (https://github.com/qihoo360/xlearning)
> License:
> 
> * Copyright (c) 2017, qihoo
> * All rights reserved.
> * Redistribution and use in source and binary forms, with or without
> * modification, are permitted provided that the following conditions are met:
> *
> *     * Redistributions of source code must retain the above copyright
> *       notice, this list of conditions and the following disclaimer.
> *     * Redistributions in binary form must reproduce the above copyright
> *       notice, this list of conditions and the following disclaimer in the
> *       documentation and/or other materials provided with the distribution.
> *     * Neither the name of qihoo nor the names of its contributors may be 
> *       used to endorse or promote products derived from this software
> *       without specific prior written permission.
> *
> * THIS SOFTWARE IS PROVIDED BY THE QIHOO AND CONTRIBUTORS "AS IS" AND ANY
> * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
> * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
> * DISCLAIMED. IN NO EVENT SHALL THE QIHOO AND CONTRIBUTORS BE LIABLE FOR ANY
> * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
> * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
> * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
> * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
> * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
> * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
> ```
So now that we're starting from scratch, I'm assuming this is no longer applicable? @erwa

erwa:
> So now that we're starting from scratch, I'm assuming this is no longer applicable? @erwa

@pdtran3k6 , yeah, I don't think the license is applicable anymore. Some of my other comments may also no longer apply. Use your best judgment.

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private HdfsConfiguration setUpHdfsConf() {
+        HdfsConfiguration conf = new HdfsConfiguration();
+
+        if (System.getenv("HADOOP_CONF_DIR") != null) {
+            conf.addResource(new Path(System.getenv(HADOOP_CONF_DIR) + File.separatorChar + CORE_SITE_CONF));
+            conf.addResource(new Path(System.getenv(HADOOP_CONF_DIR) + File.separatorChar + HDFS_SITE_CONF));
+        }
+
+        // return `kerberos` if on test cluster (Uno) / prod cluster (Holdem).
```
erwa: Don't mention the LinkedIn cluster names here -- just say "if on Kerberized cluster"

tony-history-server/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
erwa: You shouldn't need to check in another gradlew and gradle/ directory under tony-history-server. You should be able to build using the gradlew + gradle/ in the root of the TonY project. E.g.:

```
# from the root of TonY repo
./gradlew :tony-history-server:build
```

erwa:
In the future, when you're ready for re-review, please add a comment to the PR thread. Otherwise, we won't receive a notification and will assume you're still working on the changes.

tony-history-server/build.gradle
```diff
@@ -0,0 +1,47 @@
+plugins {
+    id 'play'
+    id 'idea'
+}
+
+def playVersion = "2.6.20"
+def hadoopVersion = "3.1.1"
+def scalaVersion = System.getProperty("scala.binary.version", "2.12")
```
erwa: Please define these versions in the root build.gradle file and reuse them here. We shouldn't define hadoopVersion in 2 places.

You should be able to make a variable globally accessible by defining it like so:

```
ext.hadoopVersion = "3.1.1"
...
```

Then you should be able to access it from subprojects using `ext.hadoopVersion`

tony-history-server/buildAndDeploy.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
```
erwa: Please add a comment with usage instructions. Looks like this script requires one argument.

Would be good to describe the purpose and usage of this script in the README.md you added, too.

tony-history-server/buildAndDeploy.sh
```diff
@@ -0,0 +1,24 @@
...
+echo -n "Packaging distribution zip..." >> $DEPLOY_LOG
+echo -n "Packaging distribution zip..."
+gradle dist 2>&1 >> $DEPLOY_LOG
+echo >> $DEPLOY_LOG
+echo
+
+echo -n "Copying over to $1..." >> $DEPLOY_LOG
+echo -n "Copying over to $1..."
+scp build/distributions/playBinary.zip phattran@$1:~/. 2>&1 >> $DEPLOY_LOG
```
erwa: Replace `phattran` with `$USER`

tony-history-server/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+    }
+}
+
+dependencies {
+    play "com.typesafe.play:play-guice_$scalaVersion:$playVersion"
+    play "com.typesafe.play:play-logback_$scalaVersion:$playVersion"
+    play "org.apache.hadoop:hadoop-hdfs-client:$hadoopVersion"
+    play "org.apache.hadoop:hadoop-common:$hadoopVersion"
```
erwa: Please keep alphabetized.

Also, instead of hardcoding the dependencies here, please define them in the root build.gradle file and reference that, e.g.: `deps.hadoop.common`, `deps.hadoop.hdfs_client`, etc.

tony-history-server/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+
+repositories {
+    jcenter()
+    maven {
+        name "lightbend-maven-release"
+        url "https://repo.lightbend.com/lightbend/maven-releases"
+    }
+    ivy {
+        name "lightbend-ivy-release"
+        url "https://repo.lightbend.com/lightbend/ivy-releases"
```
erwa: Are all the dependencies required to build present in our internal Artifactory? If not, we should request to import them.

conf/log4j.properties
```diff
@@ -0,0 +1,22 @@
...
+log4j.appender.console.target=System.err
+log4j.appender.console.layout=org.apache.log4j.PatternLayout
+log4j.appender.console.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n
+log4j.logger.org.apache.zookeeper=ERROR
+log4j.logger.org.apache.hadoop.hdfs.distributedavatarfilesystem=ERROR
+log4j.logger.org.apache.hadoop.util.NativeCodeLoader=ERROR
+# Settings to quiet third party logs that are too verbose
+
+# Settings the HistoryServer logs
+#log4j.logger.net.qihoo.xlearning.jobhistory=DEBUG,RFA
```
erwa: Please delete all these commented-out settings

bin/start-history-server.sh
erwa: I think we may want to reuse some of these variables/settings client-side, too. What do you think about renaming these variables/settings `TONY_HOME`, `TONY_CONF_DIR`, `TONY_CLASSPATH`, `tony-site.xml`, etc.?

tony-history-server/buildAndDeploy.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+DEPLOY_LOG=deploy.log
+
+echo -n "Cleaning up old builds..." > $DEPLOY_LOG
+echo -n "Cleaning up old builds..."
```
erwa: To avoid duplicating all these `echo` lines, you could use `tee` to write to stdout and to a file. Example:

```
echo -n "Cleaning up old builds..." | tee $DEPLOY_LOG
```

tony-history-server/conf/routes
```diff
@@ -0,0 +1,10 @@
+# Routes
+# This file defines all application routes (Higher priority routes first)
+# ~~~~
+
+# An example controller showing a sample home page
```
oliverhu: need some clean ups

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.hadoop.hdfs.HdfsConfiguration;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.yarn.api.ApplicationConstants;
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
+
+import play.Logger.ALogger;
+import play.Logger;
+import play.mvc.*;
+
+public class HomeController extends Controller {
```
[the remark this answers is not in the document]
pdtran3k6: How about IndexController?

tony-history-server/build.gradle
pdtran3k6: I'm not sure. All I know from gradle is that you have to include them. https://docs.gradle.org/current/userguide/play_plugin.html#sec:play_usage

conf/log4j.properties
pdtran3k6: This is from older commits (when we were considering forking Qihoo's repo). I'm gonna remove them in the next several commits.

bin/start-history-server.sh
pdtran3k6: https://github.com/linkedin/TonY/pull/29#discussion_r228000420

conf/log4j.properties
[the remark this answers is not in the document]
pdtran3k6: https://github.com/linkedin/TonY/pull/29#discussion_r228000420

conf/ths-site.xml
```diff
@@ -0,0 +1,57 @@
...
+    <property>
+        <name>ths.history.webapp.address</name>
+        <value>0.0.0.0:19886</value>
+    </property>
+    <property>
+        <name>ths.history.webapp.port</name>
+        <value>19886</value>
+    </property>
+    <property>
+        <name>ths.tf.board.history.dir</name>
```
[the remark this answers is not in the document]
pdtran3k6: https://github.com/linkedin/TonY/pull/29#discussion_r228000420

conf/ths-site.xml
```diff
@@ -0,0 +1,57 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License. See accompanying LICENSE file.
+-->
+
+<!-- Put site-specific property overrides in this file. -->
+
+<configuration>
```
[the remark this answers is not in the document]
pdtran3k6: https://github.com/linkedin/TonY/pull/29#discussion_r228000420

pdtran3k6:
Haven't addressed https://github.com/linkedin/TonY/pull/29#discussion_r227925765, but the rest are fixed.

tony-history-server/app/controllers/HomeController.java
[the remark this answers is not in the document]
erwa: I actually liked HomeController :). @oliverhu what name were you thinking?

tony-history-server/build.gradle
```diff
@@ -0,0 +1,41 @@
+apply plugin: "play"
+apply plugin: "idea"
+
+model {
+    components {
+        play {
```
erwa: Please use two-space indentation in this build.gradle to be consistent with our other Gradle files.

tony-history-server/buildAndDeploy.sh
```diff
@@ -0,0 +1,63 @@
...
+# 3 - Copying to remote host failed
+#
+#
+# CHANGELOG:
+# OCT 24 2018 PHAT TRAN
+############################################################################################################
+DEPLOY_LOG=deploy.log
+
+echo "Cleaning up old builds..." | tee $DEPLOY_LOG
+gradle clean 2>&1 | tee -a $DEPLOY_LOG
```
erwa: I believe this can be simplified to:

```
gradle clean |& tee -a $DEPLOY_LOG
```

tony-history-server/gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
erwa: You shouldn't need to check this in. We should only have one gradlew checked in in the root directory.

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+        // return `kerberos` if on test cluster (Uno) / prod cluster (Holdem).
+        // return `simple` if develop locally.
+        LOG.debug("Hadoop Auth Setting: " + conf.get("hadoop.security.authentication"));
+        return conf;
+    }
+
+    private void setUpKeytab(HdfsConfiguration hdfsConf) {
+        FileSystem fs = null;
+        boolean isSecurityEnabled = hdfsConf.get("hadoop.security.authentication") == "kerberos" ? true : false;
+        LOG.debug("Security: " + isSecurityEnabled);
```
erwa: I don't think this log line and the isSecurityEnabled boolean are needed, since you already log the value on line 47.

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+        if (isSecurityEnabled) {
+            try {
+                UserGroupInformation.setConfiguration(hdfsConf);
+                UserGroupInformation.loginUserFromKeytab(config.getString("keytab.user"),
+                        config.getString("keytab.location"));
+            } catch (IOException e) {
+                LOG.error("Failed to set up keytab");
```
erwa: Please include `e` in the LOG call. E.g.: `LOG.error("...", e)`

tony-history-server/build.gradle
erwa: Looks like we have the Scala 2.11 versions of play-guice and play-logback internally, but not the 2.12 ones. Please use Scala 2.11 (see my other comment).

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+        String fileContent = "";
+        String line = "";
+        BufferedReader bufReader = null;
+        Path dstPath = new Path(tonyConfigPath);
+
+        try {
+            FSDataInputStream inStrm = fs.open(dstPath);
+            bufReader = new BufferedReader(new InputStreamReader(inStrm));
+            while ((line = bufReader.readLine()) != null) {
+                fileContent += line;
```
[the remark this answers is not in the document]
pdtran3k6: Since the goal is to eventually parse the XML into a Java object to populate the html table so I won't address this issue and update it with the XML parsing code.

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+                UserGroupInformation.loginUserFromKeytab(config.getString("keytab.user"),
+                        config.getString("keytab.location"));
+            } catch (IOException e) {
+                LOG.error("Failed to set up keytab");
+            }
+        }
+        return;
+    }
+
+    private FileSystem initFs(HdfsConfiguration hdfsConf) {
```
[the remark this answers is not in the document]
pdtran3k6: How about `getInitializedFs`? 🤔

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+        try {
+            fs = FileSystem.get(hdfsConf);
+        } catch (IOException e) {
+            LOG.error("Failed to instantiate HDFS FileSystem object");
+        }
+        return fs;
+    }
+
+    private String contentOf(FileSystem fs, String tonyConfigPath) throws IOException {
+        String fileContent = "";
```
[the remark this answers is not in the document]
pdtran3k6: Noted. Also: https://github.com/linkedin/TonY/pull/29#discussion_r228280785

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+        String line = "";
```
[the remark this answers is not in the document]
pdtran3k6: Noted. Also: https://github.com/linkedin/TonY/pull/29#discussion_r228280785

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    private FileSystem initFs(HdfsConfiguration hdfsConf) {
+        FileSystem fs = null;
+        try {
+            fs = FileSystem.get(hdfsConf);
+        } catch (IOException e) {
+            LOG.error("Failed to instantiate HDFS FileSystem object");
+        }
+        return fs;
```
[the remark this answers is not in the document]
pdtran3k6: I still have to return `FileSystem` at the end after the `catch` due to function signature. I guess I don't have to make a variable `fs` and just `return null;`

tony-history-server/buildAndDeploy.sh
pdtran3k6: Do you have a reference for this? This syntax looks unfamiliar to me...

erwa: See https://stackoverflow.com/a/37085215/1128392

erwa: `|&` is just syntactic sugar for `2>&1 |`

tony-history-server/app/controllers/HomeController.java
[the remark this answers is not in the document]
erwa: That's fine, or just `getFileSystem()`

tony-history-server/app/controllers/HomeController.java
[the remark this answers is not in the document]
erwa: Ok, that's fine.

tony-history-server/app/controllers/HomeController.java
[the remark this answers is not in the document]
erwa: Ah, good point. I forgot about that. Another option is to catch the IOException and then throw a RuntimeException (wrapping the IOException). Something like:
```
try {
  return FS.get(...);
catch (IOException e) {
  throw new RuntimeException(e);
}
```

If we fail to create an FS, RuntimeException makes sense since TonY History Server can't do anything if it can't read HDFS.

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,102 @@
...
+
+		// return `kerberos` if on Kerberized cluster.
+		// return `simple` if develop locally.
+		LOG.debug("Hadoop Auth Setting: " + conf.get("hadoop.security.authentication"));
+		return conf;
+	}
+
+	private void setUpKeytab(HdfsConfiguration hdfsConf) {
+		boolean isSecurityEnabled = hdfsConf.get("hadoop.security.authentication").equals("kerberos");
+		if (isSecurityEnabled) {
```
erwa: Instead of parsing `hadoop.security.authentication` yourself, you should be able to just call `UserGroupInformation.isSecurityEnabled()`

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,102 @@
...
+				UserGroupInformation.loginUserFromKeytab(config.getString("keytab.user"), config.getString("keytab.location"));
+			} catch (IOException e) {
+				LOG.error("Failed to set up keytab", e);
+			}
+		}
+	}
+
+	private FileSystem initFs(HdfsConfiguration hdfsConf) {
+		try {
+			return FileSystem.get(hdfsConf);
```
erwa: Please use two-space indents; don't use tabs.

tony-history-server/test/controllers/BrowserTest.java
```diff
@@ -0,0 +1,26 @@
...
+import play.test.TestBrowser;
+import play.test.WithBrowser;
+
+import static org.junit.Assert.assertTrue;
+import static play.test.Helpers.*;
+
+public class BrowserTest extends WithBrowser {
+
+    protected Application provideApplication() {
+        return fakeApplication(inMemoryDatabase());
```
erwa: 2-space indent here and in the HomeControllerTest

tony-history-server/buildAndDeploy.sh
pdtran3k6: 👍

pdtran3k6: For some reasons, I couldn't use the `|&` in my script.
```
Cleaning up old builds...
./buildAndDeploy.sh: line 39: syntax error near unexpected token `&'
./buildAndDeploy.sh: line 39: `gradle clean |& tee -a $DEPLOY_LOG'
```

pdtran3k6: I'm going to revert back to the old version of the script (`2>&1` instead of `|&`) and fixed the tab/space indent issue in the next commit.

tony-history-server/app/controllers/HomeController.java
pdtran3k6: `UserGroupInformation.isSecurityEnabled()` returns `false` for both test cluster and local whereas parsing `hadoop.security.authentication` will give you `false` locally, and `true` on test cluster.

tony-history-server/app/controllers/HomeController.java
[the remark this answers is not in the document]
pdtran3k6: 👍

.gitignore
```diff
@@ -11,16 +12,21 @@
...
 .settings
 .svn
+.vscode/
 build/
 dependency-reduced-pom.xml
+log/
 out/
 target
 tony-final.xml
+tony-history-server/conf/application.conf
```
oliverhu: why we need to specifically ignore this file ? not conf/ or other files?

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.yarn.api.ApplicationConstants;
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
+
+import play.Logger.ALogger;
+import play.Logger;
+import play.mvc.*;
+
+
+public class HomeController extends Controller {
```
oliverhu: Didn't get a chance to reply to previous conversation, i'd prefer something like TonyHomePageController - this controller is for the homepage of tony

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,102 @@
...
+  private final String CORE_SITE_CONF = YarnConfiguration.CORE_SITE_CONFIGURATION_FILE;
+  private final String HDFS_SITE_CONF = "hdfs-site.xml";
+
+  @Inject
+  public HomeController(Config config) {
+    this.config = config;
+  }
+
+  private HdfsConfiguration setUpHdfsConf() {
+    HdfsConfiguration conf = new HdfsConfiguration();
```
oliverhu: why not inject hdfsconfiguration as well ?

tony-history-server/app/controllers/HomeController.java
```diff
@@ -0,0 +1,102 @@
...
+      try {
+        UserGroupInformation.setConfiguration(hdfsConf);
+        UserGroupInformation.loginUserFromKeytab(config.getString("keytab.user"), config.getString("keytab.location"));
+      } catch (IOException e) {
+        LOG.error("Failed to set up keytab", e);
+      }
+    }
+  }
+
+  private FileSystem getFileSystem(HdfsConfiguration hdfsConf) {
```
oliverhu: normally these functions which has nothing to do with current view should be moved to a Utils class as Controller class are the first ones to explode in size.

tony-history-server/app/controllers/HomeController.java
erwa: I just did a test and `UGI.isSecurityEnabled()` should return `true` (assuming core-site.xml has `hadoop.security.authentication` set to `kerberos`) as long as you call `UserGroupInformation.setConfiguration(conf);` first.

erwa: If you move line 55 (`UserGroupInformation.setConfiguration(hdfsConf);`) to the beginning of this method, then `UGI.isSecurityEnabled()` should work as expected.

tony-history-server/app/controllers/HomeController.java
erwa: What do you mean by "inject"? Use Guice to inject it? Something else?

tony-history-server/app/controllers/HomeController.java
erwa: Agreed. We can probably create a HadoopUtils class and move all these utility functions there.

tony-history-server/buildAndDeploy.sh
erwa: Most likely you have an older version of Bash installed on your Mac (me too, actually). According to the SO post I linked above, `|&` was introduced in Bash 4.0. On my Mac, I have 3.2.57, and `|&` doesn't work either. But once I upgraded bash (`brew install bash`), it worked. (Also have to change `#!/bin/bash` to `#!/usr/bin/env bash` so the new `bash` gets picked up)

tony-history-server/app/controllers/HomeController.java
erwa: That sounds good to me.

tony-history-server/app/controllers/HomeController.java
oliverhu: yah, use Guice or somethign to @Inject it, like the one above

.gitignore
pdtran3k6: Because the file contains LI HDFS credentials.

tony-history-server/app/controllers/HomeController.java
pdtran3k6: We can probably do that when the app gets more complex with more views/controllers. Right now, I think it's ok to include these in HomeController.

tony-history-server/app/controllers/HomeController.java
pdtran3k6: I'll look into it.

tony-history-server/app/controllers/HomeController.java
pdtran3k6: I think right now, we don't have enough views/controllers where naming becomes a problem just yet. We can iterate and refactor later as the app gets more complex.

tony-history-server/app/controllers/HomeController.java
oliverhu: It should not be included at HomeController.. these would be needed across different controllers next month. I'd suggest just do that in this PR.

.gitignore
oliverhu: ah ok

erwa:
PR looks good. We can address remaining feedback in a follow-up RB.

PR #38, opened 2018-10-15
Remove maven plugin

hungj:


hungj:
@erwa  it was causing build failures when adding the overrideBuildEnvironment block

Also I'm able to publish without it. So it seems redundant.

PR #39, opened 2018-10-16
Break cases for ML frameworks

hungj:


hungj:
Will address this as part of https://github.com/linkedin/TonY/pull/40

erwa:
Oh, I think doing this as a separate PR is cleaner / more disciplined. This fix is unrelated to #40 so I think it should be done separately, even if it's just 2 lines of change.

hungj:
Hmm ok, I thought so too, but when I pushed #40 I realized it also contained this change so I just went with it :)

Issue #31, opened 2018-10-05
TonY hangs if a worker is killed due to OOM


oliverhu:
TonY hangs if a worker is killed due to OOM

oliverhu:
We should have some doc/tests for how we handle different edge cases.

oliverhu:
Fixed in #43

PR #43, opened 2018-10-17
Fix TonY hangs if a worker is killed due to OOM #31

oliverhu:


oliverhu:
Cleaned up logic to only rely on container exit status for application result.. make it easier for maintenance. We should revisit if we want to support other use cases/RMs.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -331,8 +327,8 @@ public static void main(String[] args) {
         System.exit(-1);
       }
       result = am.monitor();
-      if (result || !am.jobFailed || am.amRetryCount == 0) {
-        LOG.info("Result: " + result + ", job failed: " + am.jobFailed + ", retry count: " + am.amRetryCount);
+      if (result || am.amRetryCount == 0) {
```
erwa: Can we rename the "result" variable "succeeded"?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -954,25 +944,23 @@ public void onContainersCompleted(List<ContainerStatus> completedContainers) {
           LOG.info(diagnostics);
         }
         TonyTask task = session.getTask(containerStatus.getContainerId());
+
         if (task != null) {
+          // Ignore tasks from past sessions.
+          if(task.getSessionId() != session.sessionId) {
```
erwa: Nit: add space after `if`

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
```diff
@@ -222,4 +224,27 @@ public void testAMStopsJobAfterTimeout() {
...
+
+  /**
+   * Test that makes sure if a worker is killed due to OOM, AM should stop the training (or retry).
+   * This test might hang if there is a regression in handling the OOM scenario.
+   *
+   * The reason why we use a Constants.TEST_WORKER_TERMINATED flag instead of simply requesting more memory than
+   * allocated is that Physical Memory Enforcement doesn't seem to work under MiniYARN.
+   */
+  @Test
+  public void testAMStopsJobAfterWorkerKilledByOOM() {
```
erwa: I think you should rename this test `testAMStopsJobAfterWorker0Killed` (remove the `ByOOM` part) because the test doesn't actually simulate an OOM -- it just simulates worker 0 getting killed, which could be due to any number of reasons.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -782,6 +761,19 @@ public String registerWorkerSpec(String taskId, String spec) throws IOException
         // on another node..
         LOG.info("[" + taskId + "] Received Registration for HB !!");
         hbMonitor.register(task);
+
+        // FOR TESTING
+        // Simulation of chief worker been killed.
+        if (System.getenv(Constants.TEST_WORKER_TERMINATED).equals("true") && taskId.equals(COORDINATOR_ID)) {
```
erwa: Can we move this code block into another method like `killIfTesting()`?

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
oliverhu: Yah, will update this. I was planning to simulate OOM at the beginning but PMEM enforcement doesn't seem to work under MiniYARN cluster.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -954,25 +932,23 @@ public void onContainersCompleted(List<ContainerStatus> completedContainers) {
...
         }
         TonyTask task = session.getTask(containerStatus.getContainerId());
+
         if (task != null) {
+          // Ignore tasks from past sessions.
+          if (task.getSessionId() != session.sessionId) {
+            return;
+          }
+          // Update TensorFlowSession on the state of the task.
+          session.onTaskCompleted(task.getJobName(), task.getTaskIndex(), containerStatus.getExitStatus());
```
hungj: nit: containerStatus.getExitStatus() -> exitStatus
there's another access on line 929, if you could change that as well that'd be great :)

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -806,23 +786,21 @@ public String registerWorkerSpec(String taskId, String spec) throws IOException
...
+     * This method was used to workaround an issue that the Python script finished while the container failed to
+     * close due to GPU allocation issue, which doesn't exist anymore.
+     *
+     * Discussion: A benefit of decoupling registering execution result from container exit status is that we can decouple
+     * tony from a specific resource manager's callback logic.
+     * However, this easily introduces lots of bugs since we'd have 3 places to handle failures - register call, container
+     * complete callback and heartbeat.
+     * To make things easier, we decide to go back and piggyback on container completion to dictate the execution result
+     * of a task.
+     * Still keep this RPC call here for now.
```
erwa: Can you add a TODO here and file an Issue to remove this method if we plan to just rely on container exit status going forward?

tony-core/src/main/java/com/linkedin/tony/tensorflow/TensorFlowSession.java
```diff
@@ -222,7 +220,7 @@ public synchronized TonyTask getMatchingTask(long allocationRequestId) {
       TonyTask[] tasks = entry.getValue();
       for (int i = 0; i < tasks.length; i++) {
         if (tasks[i] == null) {
-          tasks[i] = new TonyTask(jobName, String.valueOf(i));
+          tasks[i] = new TonyTask(jobName, String.valueOf(i), sessionId);
```
erwa: This method is currently named `getMatchingTask` but has a side effect of creating a `TonyTask` and updating an array. Can we rename this method `getAndAssignMatchingTask` to indicate that it has a side effect?

oliverhu: changed to `getAndInitMatchingTask`

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: Not really want to remove this method as this still provides some debugging information to see if it is a non-zero exit code from script or from container.

PR #51, opened 2018-10-23
Better handling of killed applications

hungj:
- Fix issue where HDFS resources not deleted when application killed
- Killing YARN application should fail client immediately
- Re-sort logs printed in client

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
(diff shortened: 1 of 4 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -686,12 +678,17 @@ private void cleanUp() {
...
       }
-      FileSystem fs = FileSystem.get(hdfsConf);
-      if (appResourcesPath != null && fs.exists(appResourcesPath)) {
-        fs.delete(appResourcesPath, true);
-      }
     } catch (IOException | YarnException e) {
-      LOG.error("Failed to clean up temporary files :" + appResourcesPath, e);
+      LOG.error("Failed to finish application.", e);
+    } finally {
+      try {
```
oliverhu: can we create a util method for this `try-catch` blob? This looks the same as the one above

hungj: hmm...how do we refactor a try-catch block? we'd have to pass the block as an argument to the util function.

hungj:
hmm, I added a slightly different test case, if the amRpcClient is nulled out after run then it should no longer be possible to communicate with AM.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
oliverhu: i mean having a 
```
deleteFile(path, msg) {
   try {
   } catch {
      Log.error(msg)
   }
}
```

hungj: Ah gotcha...fixed

PR #52, opened 2018-10-24
Saved TonyApp's URL (RM URL) to tony-final.xml

pdtran3k6:
Reason: Tony History Server will retrieve the url from tony-final.xml and display it in the UI.


oliverhu:
Discussed offline, we should create a new .jhist file to save the links instead of packing that into tony.xml

Issue #47, opened 2018-10-18
Fast-fail on chief worker failure


hungj:
If chief worker fails, we should immediately fail the application. (i.e. the underlying TF distributed training may hang, so TonY should just fail the application.)

Other workers failing is a separate issue. In theory if they fail the training can continue. But it can be configurable.

PR #53, opened 2018-10-25
Fix Fast-fail on chief worker failure #47 + Renaming

oliverhu:
Fix Fast-fail on chief worker failure #47 + renaming TFSession to TonySession.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
@@ -11,8 +11,8 @@ import com.linkedin.tony.rpc.ApplicationRpc;
 import com.linkedin.tony.rpc.ApplicationRpcServer;
 import com.linkedin.tony.rpc.TaskUrl;
 import com.linkedin.tony.tensorflow.TensorFlowContainerRequest;
-import com.linkedin.tony.tensorflow.TensorFlowSession;
-import com.linkedin.tony.tensorflow.TensorFlowSession.TonyTask;
+import com.linkedin.tony.tensorflow.TonySession;
+import com.linkedin.tony.tensorflow.TonySession.TonyTask;
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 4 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -938,7 +943,12 @@ public void onContainersCompleted(List<ContainerStatus> completedContainers) {
           if (task.getSessionId() != session.sessionId) {
             return;
           }
-          // Update TensorFlowSession on the state of the task.
+
+          if (task.getTaskIndex().equals("0") && task.getJobName().equals(WORKER_JOB_NAME)) {
+            chiefWorkerFailed = true;
```
hungj: Not sure I follow this part. Won't this mark it failed no matter if the chief succeeds or fails?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -281,14 +280,14 @@ private void buildTensorFlowSession() {
...
     LOG.info("Final task command: " + taskCommand);
 
-    TensorFlowSession.Builder builder = new TensorFlowSession.Builder()
+    TonySession.Builder builder = new TonySession.Builder()
         .setTaskCmd(taskCommand)
         .setVenv(pythonVenvZip)
         .setAMAddress(amHostPort)
         .setShellEnv(shellEnv)
+        .setTonyConf(tonyConf)
         .setTaskExecutorJVMArgs(tonyConf.get(TonyConfigurationKeys.TASK_EXECUTOR_JVM_OPTS,
```
hungj: since we're passing tonyConf into the builder perhaps we can remove this and extract the jvm_opts config there?

oliverhu: oh yah, missed to clean that part. We can merge this clean up with another PR.

PR #59, opened 2018-10-30
Enable TonY to localize a list of resources to local container

oliverhu:
Enable TonY to localize a list of resources to local container before starting the training job.

This is needed to make user's life easier by automatically downloading the required resources from HDFS to local container before starting a user's script.

Some cosmetic changes.

### DIFF
--- a/tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
+++ b/tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
@@ -8,7 +8,9 @@ package com.linkedin.tony.cli;
 import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.UUID;
 
 import com.linkedin.tony.TonyClient;
@@ -51,10 +53,11 @@ public class ClusterSubmitter {
       fs.mkdirs(cachedLibPath);
       fs.copyFromLocalFile(new Path(jarLocation), cachedLibPath);
 
-      String[] updatedArgs = Arrays.copyOf(args, args.length + 2);
-      updatedArgs[args.length] = "--hdfs_classpath";
-      updatedArgs[args.length + 1] = cachedLibPath.toString();
-      exitCode = TonyClient.start(updatedArgs);
+      // Insert --hdfs_classpath at the beginning to avoid confusion when user pass in wrong arguments.
+      List<String> updatedArgs = new ArrayList<>(Arrays.asList(args));
+      updatedArgs.add(0, cachedLibPath.toString());
+      updatedArgs.add(0, "--hdfs_classpath");
+      exitCode = TonyClient.start((String[])updatedArgs.toArray());
     } catch (IOException e) {
       LOG.fatal("Failed to create FileSystem: ", e);
       exitCode = -1;
--- a/tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
+++ b/tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
@@ -8,8 +8,8 @@ package com.linkedin.tony.cli;
 import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.UUID;
 
@@ -54,7 +54,7 @@ public class ClusterSubmitter {
       fs.copyFromLocalFile(new Path(jarLocation), cachedLibPath);
 
       // Insert --hdfs_classpath at the beginning to avoid confusion when user pass in wrong arguments.
-      List<String> updatedArgs = new ArrayList<>(Arrays.asList(args));
+      List<String> updatedArgs = new LinkedList<>(Arrays.asList(args));
       updatedArgs.add(0, cachedLibPath.toString());
       updatedArgs.add(0, "--hdfs_classpath");
       exitCode = TonyClient.start((String[])updatedArgs.toArray());
--- a/tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
@@ -11,6 +11,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.ServerSocket;
+import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -23,6 +24,7 @@ import org.apache.commons.cli.Options;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.yarn.api.ApplicationConstants;
 import org.apache.hadoop.yarn.api.records.ContainerId;
@@ -59,6 +61,7 @@ public class TaskExecutor {
   private String taskId;
   private int numTasks;
   private Configuration yarnConf = new Configuration();
+  private Configuration hdfsConf = new Configuration();
   private ApplicationRpcClient proxy;
   private Map<String, String> shellEnv = new HashMap<>();
   private int hbInterval;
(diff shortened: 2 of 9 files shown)


tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
```diff
@@ -226,46 +230,46 @@ private void registerExecutionResult(int exitCode, String jobName, String jobInd
     }
   }
 
-  private class Heartbeater implements Runnable {
-    int hbMissCounter = 0;
-    int numHbToMiss;
+private class Heartbeater implements Runnable {
```
erwa: I preferred the previous indentation. I think nested classes should be indented to emphasize they are contained inside another class.

tony-core/src/main/resources/tony-default.xml
```diff
@@ -164,6 +164,12 @@
     <value>0</value>
   </property>
 
+  <property>
+    <description>Resources that will be localized to the container that executes the job. Delimited by comma.</description>
+    <name>tony.worker.resources</name>
+    <value>hdfs://mycluster:9000/user/tony/resource1,hdfs://mycluster:9000/user/tony/resources2</value>
```
erwa: Should the default be an empty string? If the above is the default, won't this cause a RuntimeException since this cluster/path won't exist usually?

tony-core/src/main/java/com/linkedin/tony/Utils.java
```diff
@@ -355,4 +353,20 @@ public static void cleanupHDFSPath(Configuration hdfsConf, Path path) {
...
   }
+
+  /**
+   * Copy a list of resources delimited by comma from hdfs to local directory.
+   * @param resources a list of resources to be copied, delimited by comma.
+   * @param localDir local directory that these resources will be copied to.
+   * @param hdfsConf the configuration file for HDFS.
+   * @throws IOException exception thrown during file copies.
+   */
+  public static void localizeResources(String resources, String localDir, Configuration hdfsConf) throws IOException {
```
erwa: Instead of adding this method and having the TaskExecutors copy resources from HDFS to local disk, I think we should just rely on YARN's localization mechanism to localize resources. See `TonyClient.addLocalResources()`. I think we should just use that. If the resource is already on HDFS, then we don't need to do the copy from local to HDFS part; we just need to create a LocalResource for it.

P.S.: The variable names `scFileStatus` and `scRsrc` in TonyClient.addLocalResources are weird. What is `sc` supposed to stand for? Can you rename these variables?

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
oliverhu: hmm, I did a copy all & paste to use the format suggested by Intellij. Let me double check the indentation for this blob.

tony-core/src/main/java/com/linkedin/tony/Utils.java
oliverhu: This is a problem with that, you can't localize a directory, that is why I piggyback on the copyToLocal method.

oliverhu: Talked to @hungj , we should have two flags, one for the dependencies for each container - libraries like what we did for tony. This part will use `localResources` to benefit from the yarn file cache as well (however, this won't persist the file hierarchy)

The other one is a function pull data from hdfs to local, this is more like a convenient method for non-power users to pull data from hdfs to local

tony-core/src/main/resources/tony-default.xml
oliverhu: will update

tony-core/src/main/java/com/linkedin/tony/Utils.java
hungj: localization will work but it shouldn't be used for localizing data, I believe the container will time out if localization has not finished by a timeout (10 min default), if data is localized and data grows, then more containers will timeout.

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
```diff
@@ -172,9 +173,9 @@ protected boolean init(String[] args) throws Exception {
     amAddress = cliParser.getOptionValue("am_address", "");
     taskCommand = cliParser.getOptionValue("task_command", "exit 0");
     timeOut = tonyConf.getInt(TonyConfigurationKeys.WORKER_TIMEOUT,
-        TonyConfigurationKeys.DEFAULT_WORKER_TIMEOUT);
+                              TonyConfigurationKeys.DEFAULT_WORKER_TIMEOUT);
```
hungj: Can we keep the indentation like this? The 4 spaces thing for continuing long lines is our convention (here and everywhere else)

tony-core/src/main/java/com/linkedin/tony/Utils.java
hungj: Can we rename this function to something other than localizeResources? Don't want to confuse it with YARN localization

tony-core/src/main/resources/tony-default.xml
```diff
@@ -164,6 +164,12 @@
     <value>0</value>
   </property>
 
+  <property>
+    <description>Resources that will be localized to the container that executes the job. Delimited by comma.</description>
```
hungj: Same here, can we just say copied to the container home dir instead of localized?

tony-core/src/main/resources/tony-default.xml
hungj: we shouldn't need to specify the hdfs uri right? Haven't tested it, but since we're calling fs.copyToLocalFile it should assume src is on hdfs and dst is local?

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
oliverhu: ok.. I guess I need to update my Intellij preference. I did a copy paste for this whole file

tony-core/src/main/resources/tony-default.xml
oliverhu: yah, will update. For the issue below:
```
we shouldn't need to specify the hdfs uri right? Haven't tested it, but since we're calling fs.copyToLocalFile it should assume src is on hdfs and dst is local?
```
I want to unify this with `src_dir` flag. Basically the resource can be either local or in remote fs.

oliverhu: There seems to be bug in github that I can't reply the comment below for some reason.. thus replying here.

tony-core/src/main/java/com/linkedin/tony/Utils.java
oliverhu: Yah, the problem is for R script, I don't think they have a way to stream data from HDFS. Basically we need a way to workaround the issue to pull the data from HDFS for the local R binary to consume.

Plan to have a method `container_resources` for libraries and another `application_resources` method for the data stuff.

oliverhu: Checked with the use cases, they normally just download the same csv files (~2GB each) and use R to process from different locations.

Updated plan it support only localized_resource array.

oliverhu: Considering it is only 2GB, the time to localize should not pass 20s, which is far away from default timeout. We can revisit when user starts to abuse this flag.

erwa: Looks like there are R packages that let you read data from HDFS. They should be able to use something like rhadoop to read directly from HDFS from their R script: https://community.hortonworks.com/articles/34047/read-from-hdfs-in-r-script.html

oliverhu: ok, I'll only add localizable resources array then.

tony-core/src/main/resources/tony-default.xml
oliverhu: i did revert this change. Maybe a later refactoring rb.

tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
```diff
@@ -51,10 +53,11 @@ public static void main(String[] args) throws URISyntaxException {
       fs.mkdirs(cachedLibPath);
       fs.copyFromLocalFile(new Path(jarLocation), cachedLibPath);
 
-      String[] updatedArgs = Arrays.copyOf(args, args.length + 2);
-      updatedArgs[args.length] = "--hdfs_classpath";
-      updatedArgs[args.length + 1] = cachedLibPath.toString();
-      exitCode = TonyClient.start(updatedArgs);
+      // Insert --hdfs_classpath at the beginning to avoid confusion when user pass in wrong arguments.
+      List<String> updatedArgs = new ArrayList<>(Arrays.asList(args));
+      updatedArgs.add(0, cachedLibPath.toString());
```
hungj: Can we just use a linkedlist here?

oliverhu:
@hungj  that's the current assumption. Plan to refactor local resources like `src/` and remote `hdfs_classpath/` into a generic resources next.

PR #62, opened 2018-10-30
Tony History Server

pdtran3k6:
Added custom script for starting THS.
Refactored + Modularized code.


tony-history-server/app/controllers/JobsConfigPageController.java
```diff
@@ -0,0 +1,47 @@
...
+  @Inject
+  public JobsConfigPageController(Config config) {
+    this.config = config;
+    getInstance(config);
+  }
+
+  public Result index(String jobId) {
+    List<JobConfig> listOfConfigs = new ArrayList<>();
+    FileSystem myFs = getFs();
+    String tonyHistoryFolder = config.getString("tony.historyFolder");
```
oliverhu: make tony.historyFolder constant

tony-history-server/app/controllers/JobsConfigPageController.java
```diff
@@ -0,0 +1,47 @@
...
+    List<Path> paths = getFilePathsFromOneJob(myFs, tonyHistoryFolder, jobId, "xml");
+
+    if (paths.size() > 0) {
+      try {
+        // hardcode first item since we only have one xml file (config file)
+        listOfConfigs = parseConfig(myFs, paths.get(0));
+      } catch (Exception e) {
+        LOG.error("Failed to fetch list of configs", e);
```
oliverhu: we still return OK in this case ?

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import play.Logger;
+
+
+/**
+ * The class handles all HDFS file operations
+ */
+public class HdfsUtils {
```
oliverhu: this seems generic for all modules. We might need a component for all utils. Something like hadoop-common.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -0,0 +1,108 @@
...
+    try (FSDataInputStream inStrm = fs.open(hdfsFilePath); FileWriter fw = new FileWriter(localFileName)) {
+      IOUtils.copy(new InputStreamReader(inStrm), fw);
+    } catch (IOException e) {
+      LOG.error("Couldn't copy from file from HDFS", e);
+      return "";
+    }
+    return localFileName;
+  }
+
+  static List<Path> getValidPaths(FileStatus[] lsJobDir, Predicate<FileStatus> fn) {
```
oliverhu: We need tests for methods in this class

tony-history-server/app/utils/SecurityUtils.java
```diff
@@ -0,0 +1,82 @@
...
+  private static FileSystem getFileSystem(HdfsConfiguration hdfsConf) {
+    try {
+      return FileSystem.get(hdfsConf);
+    } catch (IOException e) {
+      LOG.error("Failed to instantiate HDFS FileSystem object", e);
+    }
+    return null;
+  }
+
+  public static FileSystem getFs() {
```
oliverhu: why is this method in SecurityUtils? Also you don't want to use shorthand for fs..

tony-history-server/app/utils/SecurityUtils.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public static SecurityUtils getInstance(Config conf) {
+    if (instance == null) {
+      instance = new SecurityUtils(conf);
+    }
+
+    return instance;
+  }
+
+  private SecurityUtils(Config appConf) {
```
oliverhu: This looks weird. Normally Utils should only have static methods.

tony-history-server/app/controllers/JobsConfigPageController.java
pdtran3k6: That's a good point. I can make an error view and redirect it to that if it fails.

tony-history-server/README.md
```diff
@@ -43,4 +43,13 @@ Before using the script, ensure that you have set execution permission (`chmod +
 $ ./buildAndDeploy.sh user hostname.test.abc.com
 ```
 
-See [script](./buildAndDeploy.sh) for more.
\ No newline at end of file
+See [script](./buildAndDeploy.sh) for more details.
+
+- To start the server after bundling with `gradle dist`:
+1. `unzip $root/build/distributions/playBinary.zip`
```
erwa: Can you rename the zip `tony-history-server`?

tony-history-server/app/utils/SecurityUtils.java
pdtran3k6: I was inspired by dr-elephant code. https://github.com/linkedin/dr-elephant/blob/master/app/com/linkedin/drelephant/security/HadoopSecurity.java

I guess the point is to have only instance of SecurityUtil for all controllers since they all use the same privilege in kerberos, so if it's already instantiated from one of the controllers, no need to make a new one, but simple just return the SecurityUtil instance from `getInstance()`.

tony-history-server/test/controllers/BrowserTest.java
```diff
@@ -14,7 +13,7 @@
 public class BrowserTest extends WithBrowser {
 
   protected Application provideApplication() {
-    return fakeApplication(ImmutableMap.of("tony.configPath", "/dummy"));
+    return fakeApplication(inMemoryDatabase());
```
erwa: I had to set `tony.configPath` for the test to get the tests to pass. Currently, they seem to be failing: https://travis-ci.org/linkedin/TonY/builds/449003339?utm_source=github_status&utm_medium=notification

tony-history-server/app/utils/SecurityUtils.java
pdtran3k6: This is just a getter so Controllers can use retrieve fs from SecurityUtils.

tony-history-server/app/controllers/JobsMetadataPageController.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Inject
+  public JobsMetadataPageController(Config config) {
+    this.config = config;
+    getInstance(config);
+  }
+
+  public Result index() {
+    FileSystem myFs = getFs();
+    LOG.info("Successfully instantiated file system");
```
erwa: This LOG message probably doesn't need to be at the `info` level. I suggest removing it or changing it to `debug`.

tony-history-server/app/controllers/JobsMetadataPageController.java
```diff
@@ -0,0 +1,43 @@
...
+
+    List<JobMetadata> listOfMetadata = new ArrayList<>();
+    String tonyHistoryFolder = config.getString("tony.historyFolder");
```
erwa: Once you've defined a `tony.historyFolder` constant (per Keqiu's comment), use that here as well.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -0,0 +1,108 @@
...
+    return Arrays.stream(lsJobDir)
+        .filter(fileStatus -> fn.test(fileStatus))
+        .map(fileStatus -> fileStatus.getPath())
+        .collect(Collectors.toList());
+  }
+
+  public static List<Path> getFilePathsFromAllJobs(FileSystem fs, String histFolder, String fileType) {
```
erwa: Please write some Javadoc for all the utility methods you added.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -0,0 +1,108 @@
...
+    FileStatus[] lsJobDir;
+
+    try {
+      lsJobDir = fs.listStatus(new Path(jobDirSb.toString()));
+    } catch (Exception e) {
+      LOG.error("Failed scanning " + jobDirSb.toString());
+      return paths;
+    }
+
+    return getValidPaths(lsJobDir, fileStatus -> fileStatus.getPath().toString().contains(fileType));
```
erwa: Instead of implementing your own fileType filter, you can just use `FileSystem.listStatus(Path f, PathFilter filter)`.

tony-history-server/README.md
pdtran3k6: Yep.

tony-history-server/test/controllers/BrowserTest.java
pdtran3k6: I can remove the tests for now if that travis build is your concern. Let me know. @erwa

tony-history-server/app/views/tableConfig.scala.html
```diff
@@ -0,0 +1,18 @@
...
+  <tr>
+    <th>Name</th>
+    <th>Value</th>
+    <th>Final</th>
+    <th>Source</th>
+  </tr>
+  @for(c <- configs) {
+    <tr>
+      <td>@c.getName()</td>
+      <td> @c.getValue</td>
```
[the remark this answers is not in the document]
pdtran3k6: Nope. Also the brackets are unnecessary (Scala). Fixed: https://github.com/linkedin/TonY/pull/62/commits/5b04e224736abf342aad557bb64012cd74995024

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: I'll look into how tests are written for a Play application and I'll add some soon.

tony-history-server/app/controllers/JobsConfigPageController.java
pdtran3k6: Something like this: https://github.com/linkedin/TonY/blob/master/tony-core/src/main/java/com/linkedin/tony/Constants.java?

tony-history-server/app/controllers/JobsMetadataPageController.java
pdtran3k6: Removed: https://github.com/linkedin/TonY/pull/62/commits/26c4ec1f972a5c8006ec1744ade806787e940265

tony-history-server/app/controllers/JobsMetadataPageController.java
pdtran3k6: Ok will do. I just wanna confirm about the format of the constants class. https://github.com/linkedin/TonY/pull/62#discussion_r229890930

tony-history-server/app/controllers/JobsConfigPageController.java
oliverhu: yup

tony-history-server/app/utils/SecurityUtils.java
oliverhu: if a method is only supposed to be used in that file, it should be private.. and for this function, it looks like generic for every file.

tony-history-server/app/utils/SecurityUtils.java
oliverhu: ok.. rename to something other than Utils :) Utils are almost always a static class with private constructors.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: I'd still wrap it around a helper function though, since we still need to convert `FileStatus[]` to `List<Path>`. I'm not sure if it helps in terms of readability and/or maintainability. Here's a gist after refactoring: https://gist.github.com/pdtran3k6/9003ae5235342541259e36c7ccef7aa4

tony-history-server/startTHS
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env sh
```
erwa: Let's add a ".sh" suffix to this file name.

tony-history-server/startTHS
```diff
@@ -0,0 +1,193 @@
...
+#
+#
+# ENVIRONMENT VARIABLES:
+# JAVA_HOME: location of Java executable
+#
+#
+# NOTES: This script should only be run when it's in the distribution zip/tar.
+# Examples: after `gradle dist` from root, unzip the zip/tar in $root/build/distributions, then run:
+# ```
+# $ $root/build/distributions/startTHS -Dhttp.port=8080 -Dconfig.resource=prod.conf
```
erwa: Looks like an extra `$` here

tony-history-server/startTHS
```diff
@@ -0,0 +1,193 @@
...
+    else
+        PRG=`dirname "$PRG"`"/$link"
+    fi
+done
+SAVED="`pwd`"
+cd "`dirname \"$PRG\"`/.." >/dev/null
+APP_HOME="`pwd -P`"
+cd "$SAVED" >/dev/null
+
+APP_NAME="playBinary"
```
erwa: Is this still accurate? Should this be `tony-history-server`?

tony-history-server/test/controllers/BrowserTest.java
erwa: I think you changed the value to `tony.historyFolder`. We should keep and fix the tests. I think you need to set a value for the `tony.historyFolder` config or else the test will fail with `com.typesafe.config.ConfigException$Missing: No configuration setting found for key 'tony'` as seen in the Travis CI build.

tony-history-server/test/controllers/JobsConfigPageControllerTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.Test;
+import play.Application;
+import play.inject.guice.GuiceApplicationBuilder;
+import play.test.WithApplication;
+
+import static org.junit.Assert.*;
+import static play.mvc.Http.Status.*;
+
+
+public class JobsConfigPageControllerTest extends WithApplication {
```
erwa: Is this the replacement for the HomeControllerTest? If so, we should migrate the tests over.

tony-history-server/app/utils/SecurityUtils.java
erwa: `getFileSystem()` seems like it belongs more in `HdfsUtils`

tony-history-server/app/utils/SecurityUtils.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
+import play.Logger;
+
+/**
+ * The class handles authentication when cluster is security enabled
+ */
+public class SecurityUtils {
+  private static final Logger.ALogger LOG = Logger.of(SecurityUtils.class);
+  private static Config appConf;
+  private static FileSystem fs;
```
erwa: This class never seems to use `fs`; it only creates and stores it. I think this should be moved out of `SecurityUtils` and only stored where it's needed. create/getting an FS should probably be done in HdfsUtils.

tony-history-server/app/utils/SecurityUtils.java
erwa: I think SecurityUtils can be all static methods. The methods should take in a `Configuration` object if needed.

tony-history-server/app/controllers/JobsMetadataPageController.java
```diff
@@ -0,0 +1,42 @@
...
+import play.Logger.ALogger;
+import play.mvc.Controller;
+import play.mvc.Result;
+
+import static utils.HdfsUtils.*;
+import static utils.ParserUtils.*;
+import static utils.SecurityUtils.*;
+
+
+public class JobsMetadataPageController extends Controller {
```
erwa: How about JobsListPageController? JobsMetadata make it sound like you are showing metadata on a particular job rather than listing all jobs.

tony-history-server/app/controllers/JobsConfigPageController.java
```diff
@@ -0,0 +1,47 @@
...
+import play.Logger.ALogger;
+import play.mvc.Controller;
+import play.mvc.Result;
+
+import static utils.HdfsUtils.*;
+import static utils.ParserUtils.*;
+import static utils.SecurityUtils.*;
+
+
+public class JobsConfigPageController extends Controller {
```
erwa: Looks like this page only shows config for a single job. Can we rename it JobConfigPageController (no 's')?

tony-history-server/app/controllers/JobsConfigPageController.java
```diff
@@ -0,0 +1,47 @@
...
+  public Result index(String jobId) {
+    List<JobConfig> listOfConfigs = new ArrayList<>();
+    FileSystem myFs = getFs();
+    String tonyHistoryFolder = config.getString("tony.historyFolder");
+    List<Path> paths = getFilePathsFromOneJob(myFs, tonyHistoryFolder, jobId, "xml");
+
+    if (paths.size() > 0) {
+      try {
+        // hardcode first item since we only have one xml file (config file)
+        listOfConfigs = parseConfig(myFs, paths.get(0));
```
erwa: If we only have one XML file, rather than get all XML files and then read the first one, why not check for and read the one XML file directly?

tony-history-server/app/utils/SecurityUtils.java
pdtran3k6: ok so I guess a better name would be `getInitializedFileSystemObj` if we want to be technical. Kinda lengthy imo, I'd say `getInitializedFs` would probably suffice.

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -0,0 +1,88 @@
...
+import org.w3c.dom.NodeList;
+import play.Logger;
+import play.libs.Json;
+
+import static utils.HdfsUtils.*;
+
+/**
+ * The class handles parsing different file format
+ */
+public class ParserUtils {
```
erwa: Create a `private ParserUtils() { }` constructor in utility classes to prevent instantiation. See _Effective Java_ Item 3.

tony-history-server/app/controllers/JobsConfigPageController.java
pdtran3k6: Hmm so in that case can I put `Constants` in utils?

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -0,0 +1,88 @@
...
+    } catch (JsonSyntaxException e) {
+      LOG.error("Couldn't parse metadata", e);
+      return jobMetadata;
+    }
+    jobMetadata.setId(jObj.get("id").textValue());
+    jobMetadata.setJobLink(jObj.get("url").textValue());
+    jobMetadata.setConfigLink("/jobs/" + jobMetadata.getId());
+    jobMetadata.setStarted(jObj.get("started").textValue());
+    jobMetadata.setCompleted(jObj.get("completed").textValue());
+    LOG.info("Successfully parsed metadata");
```
erwa: Remove or make .debug

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -0,0 +1,88 @@
...
+
+import static utils.HdfsUtils.*;
+
+/**
+ * The class handles parsing different file format
+ */
+public class ParserUtils {
+  private static final Logger.ALogger LOG = Logger.of(ParserUtils.class);
+
+  public static JobMetadata parseMetadata(FileSystem fs, Path metricsFilePath) {
```
erwa: Please add Javadoc to this and other methods. The methods in ParserUtils make assumptions such as the file being JSON or XML that should be documented.

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -0,0 +1,88 @@
...
+    LOG.info("Successfully parsed metadata");
+    return jobMetadata;
+  }
+
+  public static List<JobConfig> parseConfig(FileSystem fs, Path configFilePath) {
+    if (!isPathValid(fs, configFilePath)) {
+      return new ArrayList<>();
+    }
+
+    File configFile = new File(copyFromHdfs(fs, configFilePath));
```
erwa: You shouldn't need to copy the file from HDFS to local. DocumentBuilder has a .parse() method that takes an InputStream. You should be able to open the HDFS file and feed the InputStream into DocumentBuilder.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: Yep, will do.

tony-history-server/app/utils/HdfsUtils.java
erwa: Ok, fine with using our own predicate if you think it's more readable.

Suggest you change the `.contains(fileType)` to `endsWith(fileType)` to be more robust.

tony-history-server/app/utils/HdfsUtils.java
erwa: These utility methods should be easy to test since you can just instantiate the input objects, directly call the methods, and then assert things about the returned output.

tony-history-server/startTHS
pdtran3k6: Oh that was intentional. The first `$` is for showing that it's in a terminal/cmd line. The second one is to get the value of the variable `root`, which in this case is the path of the project.

tony-history-server/startTHS
pdtran3k6: Yeah, it should be `tony-history-server`.

tony-history-server/app/utils/SecurityUtils.java
pdtran3k6: Ok. Fixed: https://github.com/linkedin/TonY/pull/62/commits/fb5d61579eea00e50d6ab9ae6993d69a3f89eaba

tony-history-server/startTHS
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/62/commits/b5c273129ed921f6137fe822ea156375e845974b

tony-history-server/test/controllers/JobsConfigPageControllerTest.java
pdtran3k6: No its not. This one requires access to HDFS to read file, so I will look into how to create mock HDFS obj in Play.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: So what are the rules/best practices for putting something as .info vs. .debug?

tony-history-server/app/controllers/JobsConfigPageController.java
pdtran3k6: Actually, on a second thought, I don't think we should put it in Constants file, as `tony.historyFolder` is different for dev/prod mode.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: 👍

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/62/commits/3c82d8a3ef4c31f9d8c671e734d4b1853fbb1b2a

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: 👍

pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/62/commits/00bc6de64cef38b2c35f288a31d5cb99357f3eef

tony-history-server/test/controllers/BrowserTest.java
pdtran3k6: Ok, I'm going to keep the tests, and I should be able to fix the Travis build by today.

tony-history-server/startTHS
erwa: Oh. I think it would be clearer if you used
```
$ <tony-root>/build/...
```

tony-history-server/app/utils/ParserUtils.java
erwa: INFO is usually the default log level, meaning users will generally see these messages. DEBUG messages won't get shown if the log level is INFO, since DEBUG is at a higher verbosity level. Usually, you only need to enable DEBUG logging while developing / debugging.

You want to keep the number of INFO messages small so the logs aren't flooded with repetitive messages. DEBUG logging can be more verbose since you generally enable it only when you're trying to figure out a bug or implement a new feature.

tony-history-server/app/controllers/JobsConfigPageController.java
erwa: `tony.historyFolder` is the name of the parameter, which is constant. The actual value (which will be different in dev/prod) comes from the `application.conf` file (or later from `$TONY_CONF_DIR/tony-site.xml`).

tony-history-server/app/controllers/JobsMetadataPageController.java
pdtran3k6: Hence it's 'JobsMetadata': metadata of multiple jobs, not one.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: I'm assuming this would also apply for other utils class?

erwa: Yes, that's right.

tony-history-server/app/controllers/JobsMetadataPageController.java
erwa: I agree 'JobsMetadata' is an accurate name, but I think 'JobsList' more directly captures the purpose of the page -- which lists all the TonY jobs that have finished or are running. The 'List' part emphasizes that the page shows info about multiple jobs.

Minor point, so I leave it up to you to pick a name.

tony-history-server/startTHS
pdtran3k6: Oh right that looks better. Fixed: https://github.com/linkedin/TonY/pull/62/commits/9fc1d6b103ce906183051d6b89a4b75ff6c26fd9

tony-history-server/app/controllers/JobsConfigPageController.java
```diff
@@ -0,0 +1,51 @@
...
+      myFs = SecurityUtils.getInitializedFs();
+    } catch (Exception e) {
+      return internalServerError("Failed to initialize file system", e.toString());
+    }
+    String tonyHistoryFolder = config.getString("tony.historyFolder");
+    List<Path> paths = getFilePathsFromOneJob(myFs, tonyHistoryFolder, jobId, "xml");
+
+    if (paths.size() > 0) {
+      // hardcode first item since we only have one xml file (config file)
+      listOfConfigs = parseConfig(myFs, paths.get(0));
```
erwa: I think I might have already commented on this (but couldn't find my previous comment): we only expect a single XML file per job, so just look for and read that file directly rather than listing all the files and filtering on files ending with "xml".

tony-history-server/app/controllers/JobsMetadataPageController.java
```diff
@@ -0,0 +1,54 @@
...
+    String tonyHistoryFolder = config.getString("tony.historyFolder");
+    JobMetadata tmpMetadata;
+
+    for (Path f : getFilePathsFromAllJobs(myFs, tonyHistoryFolder, "json")) {
+      tmpMetadata = parseMetadata(myFs, f);
+      if (tmpMetadata == null) {
+        LOG.error("Couldn't parse " + f.toString());
+        continue;
+      }
+      listOfMetadata.add(parseMetadata(myFs, f));
```
erwa: Why do you call parseMetadata() twice (here and on line 44)? Just use tmpMetadata here instead?

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -0,0 +1,97 @@
...
+        .map(FileStatus::getPath)
+        .collect(Collectors.toList());
+  }
+
+  public static List<Path> getFilePathsFromAllJobs(FileSystem fs, String histFolder, String fileType) {
+    List<Path> paths = new ArrayList<>();
+    FileStatus[] lsHist;
+    try {
+      lsHist = fs.listStatus(new Path(histFolder));
+    } catch (Exception e) {
```
erwa: Please catch the specific exceptions thrown by listStatus() rather than doing a blanket catch of all Exception types. In general, it's best practice to catch as specific Exceptions as possible.

Same elsewhere in your code.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -0,0 +1,97 @@
...
+            fileStatus -> fileStatus.getPath().toString().endsWith(fileType));
+      } catch (Exception e) {
+        LOG.error("Failed scanning " + item.getPath().toString(), e);
+        return new ArrayList<Path>();
+      }
+    }).flatMap(List::stream).collect(Collectors.toList());
+    return paths;
+  }
+
+  public static List<Path> getFilePathsFromOneJob(FileSystem fs, String histFolder, String jobId, String fileType) {
```
erwa: This is a funny name for a method. Seems like it should be something like `getJobFiles()`, which would return a List of all the files related to that job. Doesn't seem necessary to pass a fileType in.

tony-history-server/app/utils/SecurityUtils.java
erwa: I don't think there should be a getInstance() method in SecurityUtils if it's all static methods. Like Keqiu said, generally Utils classes are singletons and there's no need to get an instance of them; you just call the static methods directly.

tony-history-server/conf/application.example.conf
```diff
@@ -1,8 +1,8 @@
...
+play.http.secret.key = "test"
 keytab {
-    user="abc@example.com"
-    location="/path/to/abc.keytab"
+  user = "abc@example.com"
+  location = "/path/to/abc.keytab"
 }
 tony {
-    configPath="/path/to/config/file.xml"
+  historyFolder = "/path/to/history/folder/"
```
erwa: BTW, in my tony-conf RB, I named this property `tony.history.directory`. Can we standardize on that? Also, in the future, I think the app setup will have to be updated to read the configuration from tony-site.xml (which might be located at `$TONY_CONF_DIR/tony-site.xml`).

tony-history-server/test/controllers/JobsConfigPageControllerTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+  @Override
+  protected Application provideApplication() {
+    return new GuiceApplicationBuilder().build();
+  }
+
+  @Test
+  public void testIndex() {
+    // TODO: Write tests
+    assertEquals(OK, OK);
```
erwa: Please write some tests

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/62/commits/7a29abf098575b18b7b82256b6603d8d28f475f8

tony-history-server/conf/application.example.conf
pdtran3k6: With that being said, we may have to parse the `tony-site.xml` and feed it to Play server class. I can modify it in `startTHS.sh` at some point in the future.

erwa: Yeah, just wanted to give you a heads up. We can make the change in a subsequent PR.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/62/commits/f84caffeeea11f73f488a6d87502659e5c22c5b0

tony-history-server/app/utils/SecurityUtils.java
pdtran3k6: Hmm no that I think about it, this module shouldn't be in utils because it's not really a util: it's a class that only needed to be created once and share by all clients (controllers). Thus, I'm moving it to its own folder called `security`, and refactored everything accordingly. Commit: https://github.com/linkedin/TonY/pull/62/commits/89bd2fdd13e2bc207eb4e0a28e4025555cdeb514. Let me know what you think @erwa @oliverhu

erwa:
I think the two main things I'd like to see before we commit this are:

* tests for the Utils classes
* Javadoc for the Utils methods

pdtran3k6:
For https://github.com/linkedin/TonY/pull/62/commits/dbc7615b5ce100cbb9edd6a1c263a4f7e0f654fe, I didn't write Javadoc for the last method, as me and @erwa discussed offline that it would be better to refactor the method to only be reading a specific file type (config file, which is always xml). Will update Javadoc after refactoring.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -24,6 +25,12 @@
 public class HdfsUtils {
   private static final Logger.ALogger LOG = Logger.of(HdfsUtils.class);
 
+  /**
+   * Check to see if a HDFS path is valid.
```
erwa: Please define what "valid" means. It looks like this method just tests for path existence, in which case, why do we need this method? Why not just call `fs.exists(filePath)` directly?

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -33,6 +40,12 @@ static boolean isPathValid(FileSystem fs, Path filePath) {
     }
   }
 
+  /**
+   * Retrieve the content of the file on HDFS.
```
erwa: I think you should mention `as a string` in the description.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -48,13 +61,26 @@ static String contentOfHdfsFile(FileSystem fs, Path filePath) {
     }
   }
 
+  /**
+   * Return all paths that satisfy the <code>fn</code> predicate.
```
erwa: It doesn't make a difference here, but in general, `{@code ...}` is preferable to using `<code>...</code>` as `{@code ...}` lets you embed `<` and `>` without escaping them as `&lt;` and `&gt;`. Ref: https://stackoverflow.com/a/1671106/1128392

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: You're correct, but imo for the sake of code modularity, we should have the method. Calling `fs.exists` directly would require the try/catch block every time you call it, which makes the caller method more verbose than it should be.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: I was considering using backticks (\`) to wrap the variable name in the doc, but I think I saw it somewhere where they use `<code></code>` as the variable's wrapper. Which one do you guys prefer? @erwa @oliverhu @hungj

tony-history-server/app/utils/HdfsUtils.java
erwa: Ah, I see. Avoiding try/catch blocks is a reasonable reason for adding this method.

Can we still update the description and replace `is valid` with `exists`? I think that's clearer. Also, I suggest renaming the method `pathExists`.

tony-history-server/app/utils/HdfsUtils.java
erwa: Backticks is non-standard and won't cause the enclosed text to get rendered as monospace when Javadoc HTML is generated. Let's stick with `{@code ...}`.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * List all file paths with a specific file type from all jobs directories.
+   * @param fs FileSystem object.
+   * @param histFolder full path of the history folder.
+   * @param fileType file extension (json, jhist, xml, etc.).
+   * @return A list of Path objects that has the same <code>fileType</code> in all job directories.
+   */
+  public static List<Path> getFilePathsFromAllJobs(FileSystem fs, String histFolder, String fileType) {
```
erwa: As discussed offline, we shouldn't need `getFilePathsFromAllJobs` and `getFilePathsFromOneJob` because we are creating the folder structure and file naming conventions. We should be able to just directly check for and read the expected file names.

tony-history-server/test/utils/TestHdfsUtils.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.ArrayList;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.junit.Test;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
```
erwa: Why are you using Mockito at all in these tests? You should be able to set up some test folders and files in a `@BeforeClass` method and then all the tests can run against those real folders/files. Using real files will make the tests easier to reason about and exercise the code more thoroughly.

erwa:
> For [dbc7615](https://github.com/linkedin/TonY/commit/dbc7615b5ce100cbb9edd6a1c263a4f7e0f654fe), I didn't write Javadoc for the last method, as me and @erwa discussed offline that it would be better to refactor the method to only be reading a specific file type (config file, which is always xml). Will update Javadoc after refactoring.

I think we don't need the method at all. We're designing the file naming conventions and folder layout, so we know what they are and can directly check for and read the expected files. (Also see my comment on HdfsUtils.java line 84.)

tony-history-server/test/utils/TestHdfsUtils.java
```diff
@@ -0,0 +1,211 @@
...
+
+
+public class TestHdfsUtils {
+  //  private final ByteArrayOutputStream outContent = new ByteArrayOutputStream();
+//  private final ByteArrayOutputStream errContent = new ByteArrayOutputStream();
+//  private final PrintStream originalOut = System.out;
+//  private final PrintStream originalErr = System.err;
```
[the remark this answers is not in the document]
pdtran3k6: It's still a WIP, it's my fault, I should have included it in the commit message.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: 👍

tony-history-server/test/utils/TestHdfsUtils.java
pdtran3k6: Ok I just talked to @oliverhu offline, I'll update the tests soon.

pdtran3k6: @erwa Fixed: https://github.com/linkedin/TonY/pull/62/commits/5080c611e482d8adbf9df48941573ef69ee0f4c2

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/62/commits/5080c611e482d8adbf9df48941573ef69ee0f4c2

oliverhu:
There has been 59 commits in this PR.. shall we clean up the existing and move forward with new features?

pdtran3k6:
@oliverhu 💯yeah I think so too... I'll make a new branch after merging this PR.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: Resolved: https://github.com/linkedin/TonY/pull/62/commits/63c80af94299a1db84d287f19e8e2c3ed91738de

tony-history-server/app/controllers/JobsMetadataPageController.java
```diff
@@ -0,0 +1,55 @@
...
+    if (myFs == null) {
+      return internalServerError("Failed to initialize file system");
+    }
+
+    List<JobMetadata> listOfMetadata = new ArrayList<>();
+    String tonyHistoryFolder = config.getString("tony.historyFolder");
+    JobMetadata tmpMetadata;
+
+    for (Path f : getMetadataFilePaths(myFs, tonyHistoryFolder)) {
+      tmpMetadata = parseMetadata(myFs, f);
```
erwa: This is going to be expensive (pounding the NameNode) to do every time someone visits the jobs list page. We definitely need to implement a cache, but we can do so in a follow-up PR.

tony-history-server/app/hadoop/Security.java
```diff
@@ -0,0 +1,44 @@
...
+  private Config appConf;
+
+  @Inject
+  public Security(Config appConf) {
+    this.appConf = appConf;
+    setUpKeytab(Configuration.getHdfsConf());
+  }
+
+  private void setUpKeytab(HdfsConfiguration hdfsConf) {
+    boolean isSecurityEnabled = hdfsConf.get("hadoop.security.authentication").equals("kerberos");
```
erwa: You can import and use `org.apache.hadoop.fs.CommonConfigurationKeysPublic.HADOOP_SECURITY_AUTHENTICATION` here instead.

I think you can also use `org.apache.hadoop.security.UserGroupInformation.AuthenticationMethod.KERBEROS.name()`

tony-history-server/app/hadoop/Security.java
```diff
@@ -0,0 +1,44 @@
...
+    if (isSecurityEnabled) {
+      try {
+        UserGroupInformation.setConfiguration(hdfsConf);
+        UserGroupInformation.loginUserFromKeytab(appConf.getString("keytab.user"),
+            appConf.getString("keytab.location"));
```
erwa: We should define "keytab.location" and "keytab.user" as `static final` class variables instead of inlining them here.

tony-history-server/test/utils/TestHdfsUtils.java
```diff
@@ -0,0 +1,116 @@
...
+    String histFolder = "./test/resources/typicalHistFolder";
+    List<Path> expectedRes = new ArrayList<>();
+    StringBuilder sb = new StringBuilder();
+
+    for (int i = 1; i < 6; ++i) {
+      sb.append("file:").append(System.getProperty("user.dir"));
+      sb.append("/test/resources/typicalHistFolder/job").append(i);
+      sb.append("/metadata.json");
+      expectedRes.add(new Path(sb.toString()));
+      sb = new StringBuilder();
```
erwa: Would recommend moving line 89 to the beginning of the for loop and removing this line

erwa:
Please do all new enhancements, feature work, and bug fixes in separate PRs (use a separate branch for each). Try to keep each PR focused on a single thing -- makes it easier to review and merge.

tony-history-server/app/controllers/JobsMetadataPageController.java
pdtran3k6: Agreed.

tony-history-server/app/hadoop/Security.java
pdtran3k6: I think we can only set the variables (keytab location and keytab user) after the constructor is called, and so we can't declare `final` on the variables. But I agree with the class variable part. We can set directly from `Config` injected from constructor without the `appConf` variable.

tony-history-server/app/hadoop/Security.java
pdtran3k6: For the first one, it just returns the string "hadoop.security.authentication". The second one, it always returns the string "KERBEROS", regardless of whether or not the Hadoop cluster has Kerberos set up. I don't think these attributes are what we want.

tony-history-server/app/hadoop/Security.java
erwa: I just meant defining variables for the string literals `"keytab.user"` and `"keytab.location"`. They could be defined at the top of this class like so:

```
private static final String KEYTAB_USER = "keytab.user";
private static final String KEYTAB_LOCATION = "keytab.location";
```

tony-history-server/app/hadoop/Security.java
erwa: I think there's a misunderstanding. What I'm suggesting is this:

```
import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.HADOOP_SECURITY_AUTHENTICATION;
import static org.apache.hadoop.security.UserGroupInformation.AuthenticationMethod.KERBEROS;
...
boolean isSecurityEnabled = hdfsConf.get(HADOOP_SECURITY_AUTHENTICATION).equals(KERBEROS.name());
```

This avoid hardcoding the string literals "hadoop.security.authentication" and "kerberos". If Hadoop ever changes the values of these string literals, we should get the changes for free when we upgrade Hadoop by relying on the string constants Hadoop defines.

PR #67, opened 2018-11-07
Run TonY with docker images

oliverhu:
Run TonY with docker images

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -330,7 +330,7 @@ public class TonyClient {
     if (tonyConf.getBoolean(TonyConfigurationKeys.DOCKER_ENABLED, TonyConfigurationKeys.DEFAULT_DOCKER_ENABLED)) {
       String imagePath = tonyConf.get(TonyConfigurationKeys.DOCKER_IMAGE);
       if (imagePath == null) {
-        LOG.error("Docker is enabled but the docker image path is not set.");
+        LOG.error("Docker is enabled but " + TonyConfigurationKeys.DOCKER_IMAGE + " is not set.");
         return false;
       } else {
         containerEnv.put(ContainerRuntimeConstants.ENV_CONTAINER_TYPE, "docker");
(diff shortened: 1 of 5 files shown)


tony-core/src/test/java/com/linkedin/tony/TestTonyConfigurationFields.java
```diff
@@ -29,6 +29,7 @@ public void initializeMemberVariables() {
     xmlPropsToSkipCompare.add(TonyConfigurationKeys.getMemoryKey(Constants.WORKER_JOB_NAME));
     xmlPropsToSkipCompare.add(TonyConfigurationKeys.getVCoresKey(Constants.WORKER_JOB_NAME));
     xmlPropsToSkipCompare.add(TonyConfigurationKeys.getGPUsKey(Constants.WORKER_JOB_NAME));
+    configurationPropsToSkipCompare.add(TonyConfigurationKeys.DOCKER_IMAGE);
```
erwa: Why do you use `configurationPropsToSkipCompare` here instead of `xmlPropsToSkipCompare`? What's the difference?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -325,6 +327,17 @@ private boolean init(String[] args) throws ParseException {
       shellEnv.putAll(Utils.parseKeyValue(envs));
     }
 
+    if (tonyConf.getBoolean(TonyConfigurationKeys.DOCKER_ENABLED, TonyConfigurationKeys.DEFAULT_DOCKER_ENABLED)) {
+      String imagePath = tonyConf.get(TonyConfigurationKeys.DOCKER_IMAGE);
+      if (imagePath == null) {
+        LOG.error("Docker is enabled but the docker image path is not set.");
```
hungj: Maybe print the conf name here as well?

tony-core/src/test/java/com/linkedin/tony/TestTonyConfigurationFields.java
hungj: I think configurationPropsToSkip ignores the confs in TonyApplicationKeys, xmlPropsToSkip ignores confs in tony-default.xml

oliverhu: The comment above is right

oliverhu:
Will revamp the [Usage] part of Readme in another PR, that grants another review :D

PR #79, opened 2018-11-13
Write config to HDFS; generate history file

pdtran3k6:
Write config.xml and history file to HDFS. 
History file name has all the necessary metadata, no App Name yet.
Also assumes that the history folder location is being fetched from tony-final.xml

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
+++ b/tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
@@ -1,0 +1,58 @@
+/**
+ * Copyright 2018 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.tony;
+
+import java.io.IOException;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
+
+
+public class HistoryFileUtils {
+  private static final Log LOG = LogFactory.getLog(HistoryFileUtils.class);
+
+  public static TonyJobMetadata createMetadataObj(Configuration yarnConf, String appId, long started, long completed,
+      boolean status) {
+    String jobStatus = status ? "SUCCEEDED" : "FAILED";
+    String url = "http://" + yarnConf.get(YarnConfiguration.RM_WEBAPP_ADDRESS) + "/cluster/app/" + appId;
+    String user = null;
+    try {
+      user = UserGroupInformation.getCurrentUser().getShortUserName();
+    } catch (IOException e) {
+      LOG.error("Failed reading from disk. Set user to null", e);
+    }
+    return new TonyJobMetadata(appId, url, started, completed, jobStatus, user);
+  }
+
+  public static void createHistoryFile(FileSystem fs, TonyJobMetadata metaObj, Path jobDir) {
+    Path historyFile = new Path(jobDir, generateFileName(metaObj));
+    try {
+      fs.create(historyFile);
+    } catch (IOException e) {
+      LOG.error("Failed to create history file", e);
+    }
+  }
+
+  private static String generateFileName(TonyJobMetadata metadataObj) {
+    StringBuilder sb = new StringBuilder();
+    sb.append(metadataObj.getId());
+    sb.append("-");
+    sb.append(metadataObj.getStarted());
+    sb.append("-");
+    sb.append(metadataObj.getCompleted());
+    sb.append("-");
+    sb.append(metadataObj.getUser());
+    sb.append("-");
+    sb.append(metadataObj.getStatus());
+    sb.append(".jhist");
+    return sb.toString();
+  }
+
+  private HistoryFileUtils() { }
+}
--- a/tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
+++ b/tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
@@ -7,31 +7,15 @@ package com.linkedin.tony;
 import java.io.IOException;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.security.UserGroupInformation;
-import org.apache.hadoop.yarn.conf.YarnConfiguration;
 
 
 public class HistoryFileUtils {
   private static final Log LOG = LogFactory.getLog(HistoryFileUtils.class);
 
-  public static TonyJobMetadata createMetadataObj(Configuration yarnConf, String appId, long started, long completed,
-      boolean status) {
-    String jobStatus = status ? "SUCCEEDED" : "FAILED";
-    String url = "http://" + yarnConf.get(YarnConfiguration.RM_WEBAPP_ADDRESS) + "/cluster/app/" + appId;
-    String user = null;
-    try {
-      user = UserGroupInformation.getCurrentUser().getShortUserName();
-    } catch (IOException e) {
-      LOG.error("Failed reading from disk. Set user to null", e);
-    }
-    return new TonyJobMetadata(appId, url, started, completed, jobStatus, user);
-  }
-
-  public static void createHistoryFile(FileSystem fs, TonyJobMetadata metaObj, Path jobDir) {
-    Path historyFile = new Path(jobDir, generateFileName(metaObj));
+  public static void createHistoryFile(FileSystem fs, TonyJobMetadata metadata, Path jobDir) {
+    Path historyFile = new Path(jobDir, generateFileName(metadata));
     try {
       fs.create(historyFile);
     } catch (IOException e) {
@@ -39,17 +23,17 @@ public class HistoryFileUtils {
     }
   }
 
-  private static String generateFileName(TonyJobMetadata metadataObj) {
+  private static String generateFileName(TonyJobMetadata metadata) {
     StringBuilder sb = new StringBuilder();
-    sb.append(metadataObj.getId());
+    sb.append(metadata.getId());
     sb.append("-");
-    sb.append(metadataObj.getStarted());
+    sb.append(metadata.getStarted());
     sb.append("-");
-    sb.append(metadataObj.getCompleted());
+    sb.append(metadata.getCompleted());
     sb.append("-");
-    sb.append(metadataObj.getUser());
+    sb.append(metadata.getUser());
     sb.append("-");
-    sb.append(metadataObj.getStatus());
+    sb.append(metadata.getStatus());
     sb.append(".jhist");
     return sb.toString();
   }
--- a/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
@@ -397,7 +397,7 @@ public class TonyApplicationMaster {
     try {
       user = UserGroupInformation.getCurrentUser().getShortUserName();
     } catch (IOException e) {
-      LOG.error("User not found. Set user to null", e);
+      LOG.error("Failed reading from disk. Set user to null", e);
       user = null;
     }
     return new Metadata(appIdString, url, started, completed, jobStatus, user);
@@ -453,9 +453,15 @@ public class TonyApplicationMaster {
       LOG.error("Exception while preparing AM", e);
       return false;
     }
-    if (!isJobDirSetup(fs, tonyHistoryFolder, appIdString) || !configFileExists(fs, jobDir)) {
+
+    try {
+      setupJobDir(fs, tonyHistoryFolder, appIdString);
+      writeConfigFile(fs, jobDir);
+    } catch (IOException e) {
+      LOG.error(e);
       return false;
     }
+
     rpcServer.start();
     hbMonitor.start();
     return true;
@@ -463,14 +469,12 @@ public class TonyApplicationMaster {
 
   /**
    * Create job directory under history folder.
-   * Side effects: {@code jobDir} global variable will be set if function finished successfully.
-   * Otherwise, it will just be an invalid Path object.
    * @param fs FileSystem object.
    * @param histFolder History folder location string.
    * @param appId Application ID string.
-   * @return true if the job directory is successfully set up. False otherwise.
+   * @throws IOException when failed to create job directory on HDFS
    */
-  public boolean isJobDirSetup(FileSystem fs, String histFolder, String appId) {
+  private void setupJobDir(FileSystem fs, String histFolder, String appId) throws IOException{
     jobDir = new Path(histFolder + "/" + appId + "/");
     try {
       if (!fs.exists(jobDir)) {
@@ -478,27 +482,23 @@ public class TonyApplicationMaster {
         fs.setPermission(jobDir, new FsPermission((short) 0770));
       }
     } catch (IOException e) {
-      LOG.error("Failed to create " + jobDir.toString(), e);
-      return false;
+      throw new IOException("Failed to create " + jobDir.toString(), e);
     }
-    return true;
   }
 
   /**
    * Generate config file in {@code jobDir} folder.
    * @param fs FileSystem object.
    * @param jobDir Path object of job directory (store all the files related to the job).
-   * @return true if config file is successfully written to {@code jobDir}. False otherwise.
+   * @throws IOException when failed to write config.xml to {@code jobDir}
    */
-  private boolean configFileExists(FileSystem fs, Path jobDir) {
+  private void writeConfigFile(FileSystem fs, Path jobDir) throws IOException {
     Path configFile = new Path(jobDir,"config.xml");
     try (FSDataOutputStream out = fs.create(configFile)) {
       tonyConf.writeXml(out);
     } catch (IOException e) {
-      LOG.error("Failed to write config to XML", e);
-      return false;
+      throw new IOException("Failed to write config to XML", e);
     }
-    return true;
   }
 
   /**
--- a/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
@@ -392,7 +392,7 @@ public class TonyApplicationMaster {
 
   private static Metadata createMetadataObj(long started, long completed, boolean status) {
     String jobStatus = status ? "SUCCEEDED" : "FAILED";
-    String url = "http://" + yarnConf.get("yarn.resourcemanager.webapp.address") + "/cluster/app/" + appIdString;
+    String url = "http://" + yarnConf.get(YarnConfiguration.RM_WEBAPP_ADDRESS) + "/cluster/app/" + appIdString;
     String user;
     try {
       user = UserGroupInformation.getCurrentUser().getShortUserName();
--- a/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
@@ -393,12 +393,11 @@ public class TonyApplicationMaster {
   private static Metadata createMetadataObj(long started, long completed, boolean status) {
     String jobStatus = status ? "SUCCEEDED" : "FAILED";
     String url = "http://" + yarnConf.get(YarnConfiguration.RM_WEBAPP_ADDRESS) + "/cluster/app/" + appIdString;
-    String user;
+    String user = null;
     try {
       user = UserGroupInformation.getCurrentUser().getShortUserName();
     } catch (IOException e) {
       LOG.error("Failed reading from disk. Set user to null", e);
-      user = null;
     }
     return new Metadata(appIdString, url, started, completed, jobStatus, user);
   }
--- a/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
@@ -361,7 +361,7 @@ public class TonyApplicationMaster {
     }
 
     // By this time jobDir should have been set
-    Utils.createHistoryFile(fs, Utils.createMetadataObj(yarnConf, appIdString, started, completed, succeeded), jobDir);
+    HistoryFileUtils.createHistoryFile(fs, HistoryFileUtils.createMetadataObj(yarnConf, appIdString, started, completed, succeeded), jobDir);
 
     if (succeeded) {
       LOG.info("Application Master completed successfully. exiting");
--- a/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
@@ -409,7 +409,7 @@ public class TonyApplicationMaster {
     try {
       fs.create(historyFile);
     } catch (IOException e) {
-        LOG.error("Failed to create history file", e);
+      LOG.error("Failed to create history file", e);
     }
   }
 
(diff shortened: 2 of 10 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -124,7 +151,7 @@
   private TonySession.Builder sessionBuilder;
 
   /** Configuration **/
-  private Configuration yarnConf;
+  private static Configuration yarnConf;
```
erwa: Why did you make `yarnConf` `static`?

tony-core/src/main/java/com/linkedin/tony/Utils.java
```diff
@@ -404,4 +403,8 @@ public static void addResource(String path,
       LOG.error("Failed to add " + path + " to local resources.", exception);
     }
   }
+
+  public static String generateFileName(TonyApplicationMaster.Metadata obj) {
```
erwa: I suggest a more descriptive name for the parameter such as `metadata` rather than `obj`.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -295,6 +332,8 @@ private void buildTensorFlowSession() {
    */
   public static void main(String[] args) {
     boolean succeeded = false;
+    long started;
+    long completed;
```
erwa: I addition to start time (which is when this code runs), it would be informative to know the _submit_ time as well, which is when YARN received the AM container request. I think you should be able to get that from `amRMClient.getStartTime()`

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -346,6 +390,29 @@ public static void main(String[] args) {
     }
   }
 
+  private static Metadata createMetadataObj(long started, long completed, boolean status) {
+    String jobStatus = status ? "SUCCEEDED" : "FAILED";
+    String url = "http://" + yarnConf.get("yarn.resourcemanager.webapp.address") + "/cluster/app/" + appIdString;
```
erwa: You can use `YarnConfiguration.RM_WEBAPP_ADDRESS` instead of hardcoding `"yarn.resourcemanager.webapp.address"`.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -346,6 +390,29 @@ public static void main(String[] args) {
...
+    return new Metadata(appIdString, url, started, completed, jobStatus, user);
+  }
+
+
+  private static void createHistoryFile(FileSystem fs, Metadata metaObj, Path jobDir) {
+    Path historyFile = new Path(jobDir, generateFileName(metaObj));
+    try {
+      fs.create(historyFile);
+    } catch (IOException e) {
+        LOG.error("Failed to create history file", e);
```
erwa: Unindent two spaces

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -386,6 +453,28 @@ private boolean prepare() {
       LOG.error("Exception while preparing AM", e);
       return false;
     }
+
+    // Write config file to HDFS
+    jobDir = new Path(tonyHistoryFolder + appIdString + "/");
+    try {
+      if (!fs.exists(jobDir)) {
+        fs.mkdirs(jobDir);
+        fs.setPermission(jobDir, new FsPermission((short) 0770));
```
erwa: Can you move this history folder setup logic out into a separate method like `setupHistoryDirectory` for readability?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -386,6 +453,28 @@ private boolean prepare() {
       LOG.error("Exception while preparing AM", e);
       return false;
     }
+
+    // Write config file to HDFS
+    jobDir = new Path(tonyHistoryFolder + appIdString + "/");
```
erwa: For robustness, can you add a `"/"` before appIdString in case `tonyHistoryFolder` doesn't have a trailing slash? An extra slash still works (e.g.: `hdfs dfs -ls /user//phattran` works the same as `hdfs dfs -ls /user/phattran`).

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -386,6 +453,28 @@ private boolean prepare() {
...
+      }
+    } catch (IOException e) {
+      LOG.error("Failed to create " + jobDir.toString(), e);
+      return false;
+    }
+
+    Path configFile = new Path(jobDir,"config.xml");
+
+    try (FSDataOutputStream out = fs.create(configFile)) {
+      tonyConf.writeXml(out);
```
erwa: Again, let's move this out into a separate method that clearly describes what's being done, like `writeConfigHistoryFile`.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: what?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -74,15 +77,39 @@
...
 
 public class TonyApplicationMaster {
   private static final Log LOG = LogFactory.getLog(TonyApplicationMaster.class);
 
+  /**
+   * Metadata of jobs
+   */
   private ApplicationAttemptId appAttemptID = null;
-  private String appIdString;
+  private static String appIdString;
```
oliverhu: why makes this static?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -331,12 +371,16 @@ public static void main(String[] args) {
...
     am.stop();
+    completed = System.currentTimeMillis();
     am.printTaskUrls();
     if (exitOnError) {
       LOG.fatal("Error running TonyApplicationMaster !!");
       System.exit(-1);
     }
 
+    // By this time jobDir should have been set
+    createHistoryFile(fs, createMetadataObj(started, completed, succeeded), jobDir);
```
oliverhu: Are we still trying to write to this file if this call failed?

tony-core/src/main/resources/tony-default.xml
```diff
@@ -202,4 +202,13 @@
     <name>tony.application.yarn-conf-path</name>
   </property>
 
+  <!-- History folder configurations -->
+  <property>
+    <description>
+      Location for Tony AM to write config and events on HDFS
+    </description>
+    <name>tony.history.location</name>
+    <value>/system/tony-history/</value>
```
erwa: I prefer leaving off the final `/` (so `/system/tony-history` rather than `/system/tony-history/`). If you look at other path-related configs in `$HADOOP_CONF_DIR`, you'll see that most leave off the final `/`.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -74,15 +77,39 @@
...
+   */
   private ApplicationAttemptId appAttemptID = null;
-  private String appIdString;
+  private static String appIdString;
   private String amHostPort;
+  private static FileSystem fs;
+  private static String tonyHistoryFolder;
+  private static Path jobDir;
+
+  public static class Metadata {
```
oliverhu: any reason to put this as a nested class instead of a new class?

tony-core/src/main/java/com/linkedin/tony/Utils.java
oliverhu: obj is not a good name..

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -346,6 +390,29 @@ public static void main(String[] args) {
     }
   }
 
+  private static Metadata createMetadataObj(long started, long completed, boolean status) {
```
oliverhu: move static methods to Utils class.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -346,6 +390,29 @@ public static void main(String[] args) {
...
+    String jobStatus = status ? "SUCCEEDED" : "FAILED";
+    String url = "http://" + yarnConf.get("yarn.resourcemanager.webapp.address") + "/cluster/app/" + appIdString;
+    String user;
+    try {
+      user = UserGroupInformation.getCurrentUser().getShortUserName();
+    } catch (IOException e) {
+      LOG.error("User not found. Set user to null", e);
+      user = null;
```
oliverhu: Return NULL instead of creating an invalid object?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Oh it was because calling it inside `createMetadataObj` and that was being called in `main` and `main` is static.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: I'm assuming that info (submit time) would be on the job listing page as well?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Wdym by invalid object? If the user couldn't be fetched, you would still want the history file to be created, hence I make it null. It just won't show up on the job listing page.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Noted. Fixed: https://github.com/linkedin/TonY/pull/79/commits/b67ee52d3644b2c199f8257c496395295f92b1ab

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: I'd say no, but I'm not sure how to alert users when this method fails because they might check the history server and not see their job displayed. Maybe I can just have it retry creating the file for a certain number of times?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: ?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: I was inspired by ask-hadoop-fetcher project. My reasons were that it's too trivial to make a separate class, and only TonyAM was interacting with this class at that point. But now since https://github.com/linkedin/TonY/pull/79/files/d28348e16791732b06f424d33f8a0ff8638ba51b#r233199638, we can refactor this to a separate class.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: meant why static

pdtran3k6: I think this could be reverted back to non-static once I moved all the methods that handles history files to `Utils`.

erwa: I think createHistoryFile() should be a non-static method in TonyApplicationMaster, meaning it should be called as `am.createHistoryFile()`.

`yarnConf` shouldn't be static because it should be associated with an instance of TonyApplicationMaster rather than be a single global class variable.

erwa:
Yes, I think it would be informative to show the submit time on the job listing page as well. It let's us calculate the delay between submit time and start time.


________________________________

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: By it I mean `user` attribute. The job will still show up, with the user being displayed as `null`.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/79/commits/e489758601fce6c775bd5d11e0f97289e0664a81

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/79/commits/e489758601fce6c775bd5d11e0f97289e0664a81

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/79/commits/e489758601fce6c775bd5d11e0f97289e0664a81

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: ok, then you can just init `user = null ` and then you don't need to set it here. looks weird.

tony-core/src/main/java/com/linkedin/tony/Utils.java
pdtran3k6: I was thinking about `metadataObj`, but I feel like any other name would add verbosity without additional value in terms of readability because the object type already conveys the attribute of this object. Do you have any other names in mind?

pdtran3k6: See https://github.com/linkedin/TonY/pull/79#discussion_r233593949.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Yeah that's a good point. Will do.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: 👍

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Once I do https://github.com/linkedin/TonY/pull/79#discussion_r233594390, then this shouldn't be static.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: See https://github.com/linkedin/TonY/pull/79#discussion_r233597548.

tony-core/src/main/java/com/linkedin/tony/Utils.java
erwa: Yeah, `metadataObj` is fine. I think it's still helpful to use a descriptive name despite the type since in the method body, you only see the variable name and not the type.

tony-core/src/main/resources/tony-default.xml
pdtran3k6: Ok sure. Fixed: https://github.com/linkedin/TonY/pull/79/commits/e489758601fce6c775bd5d11e0f97289e0664a81

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/79/commits/e489758601fce6c775bd5d11e0f97289e0664a81

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
erwa: (pasting reply here from below)

Yes, I think it would be informative to show the submit time on the job listing page as well. It let's us calculate the delay between submit time and start time.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: You're right. Fixed: https://github.com/linkedin/TonY/pull/79/commits/3250d31b3f8d01bf283f62e64e92a79a1519dc8e

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: `ask-hadoop-fetcher ` was a purely hackday project.. not a good reference..

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/79/commits/6ba7908b4d3312f24a810d97f8a82fa46e6c6a8e.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Yep, fixed: https://github.com/linkedin/TonY/pull/79/commits/6ba7908b4d3312f24a810d97f8a82fa46e6c6a8e.

tony-core/src/main/resources/tony-default.xml
pdtran3k6: Ok, sure. Fixed: https://github.com/linkedin/TonY/pull/79/commits/24b15e952b42cfafb8ed4d3a2bc72afb4fe3e185

tony-core/src/main/java/com/linkedin/tony/Utils.java
pdtran3k6: Ok sure. Fixed: https://github.com/linkedin/TonY/pull/79/commits/25e79cf0094f5ed38caa4ef76db2addc04b540d0

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Ok I'll put it on the back burner for the future PRs.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Yeah, I agreed. I think it should be the same for `createMetadataObj` and `generateFileName` as these methods are specifically for creating the history file. Wdyt @erwa @oliverhu?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: Actually, on a second thought, see https://github.com/linkedin/TonY/pull/79#discussion_r233588794.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
pdtran3k6: After discussing offline with @oliverhu, we've decided to move these methods to a separate class `HistoryUtils` because TonyApplicationMaster is getting quite large, and it would be hard to maintain in the future. 
Eventually, we should move all utils (`Utils`, `ProtoUtils`, `HistoryUtils`, etc.) to a `Utils` package, but that could pushed to future PRs.

pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/79/commits/16dbc9b453a5eeb6601bffd6bb5877909a329114.

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
+
+
+public class HistoryFileUtils {
+  private static final Log LOG = LogFactory.getLog(HistoryFileUtils.class);
+
+  public static TonyJobMetadata createMetadataObj(Configuration yarnConf, String appId, long started, long completed,
```
oliverhu: This method should actually be inside `TonyJobMetadata ` class, like `newInstance(a, b, c)` 
Reference: http://www.informit.com/articles/article.aspx?p=1216151

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
```diff
@@ -0,0 +1,58 @@
...
+      boolean status) {
+    String jobStatus = status ? "SUCCEEDED" : "FAILED";
+    String url = "http://" + yarnConf.get(YarnConfiguration.RM_WEBAPP_ADDRESS) + "/cluster/app/" + appId;
```
oliverhu: make this a method
```
buildRMUrl(appId) {
return "http://" + yarnConf.get(YarnConfiguration.RM_WEBAPP_ADDRESS) + "/cluster/app/" + appId;
}
```

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
```diff
@@ -0,0 +1,58 @@
...
+  public static void createHistoryFile(FileSystem fs, TonyJobMetadata metaObj, Path jobDir) {
+    Path historyFile = new Path(jobDir, generateFileName(metaObj));
+    try {
+      fs.create(historyFile);
+    } catch (IOException e) {
+      LOG.error("Failed to create history file", e);
+    }
+  }
+
+  private static String generateFileName(TonyJobMetadata metadataObj) {
```
oliverhu: I don't like `Obj` as a naming suffix, just remove that?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: Presume this is fixed in the next PR.

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
pdtran3k6: So a constructor?

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
pdtran3k6: If that's the case then I think this belongs in the generic `Utils` since it's not very specific to generating history file and might be useful for TonyClient. 
```Java
private void logTrackingAndRMUrls(ApplicationReport report) {
    LOG.info("URL to track running application (will proxy to TensorBoard once it has started): "
             + report.getTrackingUrl());
    LOG.info("ResourceManager web address for application: "
        + String.format(RM_APP_URL_TEMPLATE,
        yarnConf.get(YarnConfiguration.RM_WEBAPP_ADDRESS),
        report.getApplicationId())); // refactor this line to use `buildRMUrl(appId)`
  }
```

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
```diff
@@ -0,0 +1,58 @@
...
+    String user = null;
+    try {
+      user = UserGroupInformation.getCurrentUser().getShortUserName();
+    } catch (IOException e) {
+      LOG.error("Failed reading from disk. Set user to null", e);
+    }
+    return new TonyJobMetadata(appId, url, started, completed, jobStatus, user);
+  }
+
+  public static void createHistoryFile(FileSystem fs, TonyJobMetadata metaObj, Path jobDir) {
```
pdtran3k6: Same thing for here @oliverhu? https://github.com/linkedin/TonY/pull/79/files#r234001041

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
pdtran3k6: See https://github.com/linkedin/TonY/pull/79#discussion_r234042215.

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
pdtran3k6: Something like this: https://gist.github.com/pdtran3k6/95b2c1f31e0db270cdaa3141b93bf7ee?

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
oliverhu: Yah, looks better to be in Utils class

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
oliverhu: yes

tony-core/src/main/java/com/linkedin/tony/HistoryFileUtils.java
oliverhu: that works

PR #80, opened 2018-11-13
Add test coverage for TonyCli module and refactor TonyClient constructors

oliverhu:
There was an issue #74 that was because we broke ClusterSubmitter but we didn't have unit test for that. The primary goal of this PR is to add test coverage for that.
- Added mokito to mock out tony-core to make tony-cli testing easier.
- Refactored TonyClient contructors to make testing easier and code easier to understand.
- Clean up tests.


tony-cli/src/test/java/com/linkedin/tony/TestClusterSubmitter.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class TestClusterSubmitter {
+
+  @Test
+  public void testClusterSubmmiter() throws  Exception {
+    TonyClient client = spy(new TonyClient());
+    doReturn(0).when(client).start(); // Don't really call start() method.
+
+    ClusterSubmitter submitter = new ClusterSubmitter(client);
+    submitter.submit(new String[] {"--src_dir", "src"});
```
hungj: we need to validate the exit code right? I don't think this tests anything yet

build.gradle
```diff
@@ -65,7 +66,8 @@ ext.deps = [
     "sshd": "org.apache.sshd:sshd-core:1.1.0",
     "testng": "org.testng:testng:6.4",
     "text": "org.apache.commons:commons-text:1.4",
-    "zip4j": "net.lingala.zip4j:zip4j:1.3.2"
+    "zip4j": "net.lingala.zip4j:zip4j:1.3.2",
+    "mokito": "org.mockito:mockito-core:2.23.0"
```
hungj: mokito->mockito

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -678,53 +678,21 @@ protected ApplicationRpcClient getAMRpcClient() {
...
+  public void close() throws IOException, YarnException {
+    if (amRpcClient != null) {
+      amRpcClient.finishApplication();
     }
-    return client;
-  }
-
-  public static int start(String[] args) {
-    return start(args, new Configuration(false));
+    Utils.cleanupHDFSPath(hdfsConf, appResourcesPath);
```
hungj: I don't think we can refactor it like this? if finishApplication throws exception then appResourcesPath will not be cleaned up

oliverhu: That is right. Will update this.

tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
```diff
@@ -33,12 +34,18 @@
...
-  private ClusterSubmitter() { }
+  private ClusterSubmitter() {
+    this.client = new TonyClient(new Configuration());
+  }
+  public ClusterSubmitter(TonyClient client) {
+    this.client = client;
+  }
 
-  public static void main(String[] args) throws URISyntaxException {
+  public int submit(String[] args) throws ParseException, URISyntaxException {
```
erwa: Nit: Can we annotate these implementation methods with `@Override`. Same in NotebookSubmitter.java.

tony-cli/src/main/java/com/linkedin/tony/cli/NotebookSubmitter.java
```diff
@@ -82,17 +90,12 @@ public static void main(String[] args) throws  Exception {
...
       System.exit(-1);
     }
-    Thread clientThread = new Thread(() -> {
-      try {
-        client.run();
-      } catch (IOException | URISyntaxException | YarnException | InterruptedException e) {
-        LOG.error(e);
-      }
-    });
+    Thread clientThread = new Thread(client::start);
```
erwa: Is `client.start()` being called twice? Once here and once on line 94? I think it only needs to be called here (triggered when the `Thread` is started)?

tony-cli/src/test/java/com/linkedin/tony/TestClusterSubmitter.java
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright 2018 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.tony;
+
+import com.linkedin.tony.cli.ClusterSubmitter;
+import org.testng.annotations.Test;
+
+import static junit.framework.Assert.*;
```
erwa: Let's not mix both junit and testng in the same test. It looks like the rest of TonY uses testng, so let's stick with that.

PR #81, opened 2018-11-14
2_New feature: Added support for traversing nested folders

pdtran3k6:


pdtran3k6:
Failed tests will be fixed once https://github.com/linkedin/TonY/pull/79 is merged to keep the number of commits in this PR low.

tony-history-server/test/utils/TestHdfsUtils.java
```diff
@@ -76,38 +77,106 @@ public void testContentOfHdfsFile_throwsException() throws IOException {
...
+    sb.append("file:").append(System.getProperty("user.dir"));
+    sb.append("/test/resources/nestedHistFolder/");
+    String baseDir = sb.toString();
+
+    expectedRes.add(new Path(baseDir, "2018/01/02/job0"));
+    expectedRes.add(new Path(baseDir,"2018/01/01/job1"));
+    expectedRes.add(new Path(baseDir,"2018/12/31/job2"));
+    expectedRes.add(new Path(baseDir,"2017/07/job3"));
+    expectedRes.add(new Path(baseDir,"2017/07/job4"));
+    expectedRes.add(new Path(baseDir,"2017/07/job5"));
```
erwa: Nit: please add space after `baseDir,`

(We should enable Checkstyle on our code. Ref: https://docs.gradle.org/current/userguide/checkstyle_plugin.html)

tony-history-server/test/utils/TestHdfsUtils.java
```diff
@@ -76,38 +77,106 @@ public void testContentOfHdfsFile_throwsException() throws IOException {
...
     FileSystem mockFs = mock(FileSystem.class);
+    String regex = "job*";
 
-    when(mockFs.listStatus(new Path(histFolder))).thenThrow(new IOException("IO Excpt"));
-    List<Path> actualRes = HdfsUtils.getMetadataFilePaths(mockFs, histFolder);
+    when(mockFs.listStatus(histFolder)).thenThrow(new IOException("IO Excpt"));
+    List<Path> actualRes = HdfsUtils.getJobFolders(mockFs, histFolder, regex);
 
     assertEquals(new ArrayList<Path>(), actualRes);
     assertEquals(0, actualRes.size());
```
erwa: Having both the line 181 and 182 checks seems redundant. I think the size 0 check on line 182 is sufficient. Also, it's a little weird to be asserting something about an implementation deal (that the returned List is specifically an `ArrayList`) when the getJobFolders() method API just says it returns a `List`.

tony-history-server/app/controllers/JobsMetadataPageController.java
```diff
@@ -38,13 +39,14 @@ public Result index() {
     }
 
     List<JobMetadata> listOfMetadata = new ArrayList<>();
-    String tonyHistoryFolder = config.getString("tony.historyFolder");
+    Path tonyHistoryFolder = new Path(config.getString("tony.historyFolder"));
```
erwa: Please use `TonyConfigurationKeys.TONY_HISTORY_LOCATION`. Same elsewhere in the code. This also helps remove inconsistencies such as using `tony.historyFolder` in some places and `tony.history.location` in others.

tony-history-server/app/controllers/JobsMetadataPageController.java
```diff
@@ -38,13 +39,14 @@ public Result index() {
...
+    String jobFolderRegex = "^application_\\d+_\\d+$";
```
erwa: Please define this as a `static final String` class variable. The naming convention for such variables is generally `UPPERCASE_UNDERSCORE_SEPARATED`.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -63,27 +59,52 @@ static String contentOfHdfsFile(FileSystem fs, Path filePath) {
   }
 
   /**
-   * List all metadata file paths in {@code histFolder}.
+   * Extract the job id portion from the file path string.
+   * @param path file path string.
+   * @return job id
+   */
+  public static String getJobId(String path) {
```
erwa: The name of this method could be made more general, given what it's doing. The method is just returning the last component of a path.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -63,27 +59,52 @@ static String contentOfHdfsFile(FileSystem fs, Path filePath) {
...
+    if (path.length() == 0) return "";
```
erwa: This check is redundant as the following lines do the same thing.

I think you should add a check (and a test) for `path == null` though.

Finally, please use braces, even for one-line `if` bodies, i.e.:
```
if (...) {
  ...
}
```

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -30,28 +32,52 @@
...
+      // this should never happen unless user rename the history file
+      LOG.error("Metadata isn't valid");
+      return jobMetadata;
+    }
+
+    jobMetadata.setId(metadata[0]);
+    jobMetadata.setJobLink("/jobs/" + jobMetadata.getId());
+    jobMetadata.setConfigLink("/config/" + jobMetadata.getId());
+    jobMetadata.setStarted(simple.format(new Date(Long.parseLong(metadata[1]))));
+    jobMetadata.setCompleted(simple.format(new Date(Long.parseLong(metadata[2]))));
```
erwa: I think JobMetadata should store `long` epoch timestamps rather than Strings. At UI rendering time, we can decide what format / timezone to show the time in. I think this is converting the timestamps to a final output representation too early.

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -30,28 +32,52 @@
...
+    }
+
     try {
-      jObj = Json.parse(fileContent);
-    } catch (JsonSyntaxException e) {
-      LOG.error("Couldn't parse metadata", e);
+      metadata = Arrays.stream(fs.listStatus(jobFolderPath))
+          .filter(f -> f.getPath().toString().contains("jhist")) // ignore config.xml
+          .map(histFilePath -> HdfsUtils.getJobId(histFilePath.toString()))
+          .map(histFile -> histFile.substring(0, histFile.lastIndexOf('.'))) // remove .jhist
```
erwa: This chain of stream operations seems weird. You know the jhist file is going to be named `<appId>...jhist` so just directly look for that file. There should also be exactly 1 jhist file, so you could assert that, too. (The code currently still works if there are multiple file matches but that could lead to unexpected behavior later when you try to parse the metadata from the file name.)

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -30,28 +32,52 @@
 public class ParserUtils {
   private static final Logger.ALogger LOG = Logger.of(ParserUtils.class);
 
-  public static JobMetadata parseMetadata(FileSystem fs, Path metadataFilePath) {
-    if (!pathExists(fs, metadataFilePath)) {
-      return new JobMetadata();
+  @VisibleForTesting
+  static boolean isValidMetadata(String[] metadata, String jobIdRegex) {
```
erwa: This method is hard to understand as is. I suggest renaming it to `isValidJobHistoryFileName` and pass in the full job history file name directly.

Some Javadoc with an example would also be helpful.

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -30,28 +32,52 @@
...
+    if (!isValidMetadata(metadata, jobIdRegex)) {
+      // this should never happen unless user rename the history file
+      LOG.error("Metadata isn't valid");
+      return jobMetadata;
+    }
+
+    jobMetadata.setId(metadata[0]);
+    jobMetadata.setJobLink("/jobs/" + jobMetadata.getId());
+    jobMetadata.setConfigLink("/config/" + jobMetadata.getId());
+    jobMetadata.setStarted(simple.format(new Date(Long.parseLong(metadata[1]))));
```
erwa: Let's create a helper method `createJobMetadata(String jobHistoryFileName)`. This method can internally call `isValidMetadata` (or the new method name).

tony-history-server/test/utils/TestHdfsUtils.java
```diff
@@ -76,38 +77,106 @@ public void testContentOfHdfsFile_throwsException() throws IOException {
...
+    expectedRes.add(new Path(baseDir,"2018/12/31/job2"));
+    expectedRes.add(new Path(baseDir,"2017/07/job3"));
+    expectedRes.add(new Path(baseDir,"2017/07/job4"));
+    expectedRes.add(new Path(baseDir,"2017/07/job5"));
+
+    List<Path> actualRes = HdfsUtils.getJobFolders(fs, histFolder, regex);
+    Collections.sort(actualRes, (o1, o2) -> {
+      String job1 = HdfsUtils.getJobId(o1.toString());
+      String job2 = HdfsUtils.getJobId(o2.toString());
+      return job1.charAt(job1.length()-1) - job2.charAt(job2.length()-1);
```
erwa: I believe Checkstyle would also enforce spaces around each `-`.

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -63,27 +59,52 @@ static String contentOfHdfsFile(FileSystem fs, Path filePath) {
...
+   * @param regex regular expression string
+   * @return true if path {@code p} is a job folder path, false otherwise.
+   */
+  @VisibleForTesting
+  static boolean isJobFolder(Path p, String regex) {
+    return getJobId(p.toString()).matches(regex);
+  }
+
+  /**
+   * Find all job folders under {@code curr} that matches {@code regex} pattern
```
erwa: It'd be helpful to mention that this method is recursive

tony-history-server/app/utils/HdfsUtils.java
```diff
@@ -63,27 +59,52 @@ static String contentOfHdfsFile(FileSystem fs, Path filePath) {
...
     } catch (IOException e) {
-      LOG.error("Failed to read history folder", e);
-      return paths;
-    }
-
-    for (FileStatus nodes : lsHist) {
-      if (nodes.isDirectory()) {
-        paths.add(new Path(nodes.getPath().toString() + "/metadata.json"));
-      }
+      LOG.error("Failed to traverse down history folder");
```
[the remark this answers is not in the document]
pdtran3k6: I think I just want to return the `intermediateFolders`, assuming that it would just be an empty list when `fs.listStatus` throws error. If my assumption is incorrect, then I'll just return an empty list.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: I think I refactored this in a different branch (https://github.com/pdtran3k6/TonY/tree/phattran_jobDataCache) where I have a `newInstance` method in `JobMetadata` class (similar to the `TonyJobMetadata` in `tony-core`).

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: Sure 👍

tony-history-server/test/utils/TestHdfsUtils.java
pdtran3k6: 👍 We can do this in the following PRs. Right now, I'm just gonna manually fix these.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: See https://github.com/linkedin/TonY/pull/81/commits/6ff6714324def1e385f5995d76f389a48c1eaf15 for the updated implementation of `JobMetadata`. With this being the new implementation, do you think this comment still apply?

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: I don't think we can directly look for that file since the file name depends on other variables (`appId-started-completed-user-status.jhist`) so we can directly look for the file. Also, I was assuming there's just 1 jhist file in the `jobFolderPath`, maybe I should put that in the Javadoc for this function.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: I agree with your suggestion for the function name; however, I think we still have to do the same thing once we have the history file name: splitting and storing them in an array and check each element of the array to see if it is valid. I think it's hard to understand because of the way I format the code. How about something like this: https://gist.github.com/pdtran3k6/3b427fc1337c58a494c9b161c58c98c6

tony-history-server/test/utils/TestHdfsUtils.java
pdtran3k6: See https://github.com/linkedin/TonY/pull/81#discussion_r234429024.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/81/commits/108fe8650f76b5d9d5ae71188f3b95d170b54a93.

tony-history-server/app/controllers/JobsMetadataPageController.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/81/commits/108fe8650f76b5d9d5ae71188f3b95d170b54a93.

tony-history-server/test/utils/TestHdfsUtils.java
pdtran3k6: Ok that's fair. Fixed: https://github.com/linkedin/TonY/pull/81/commits/108fe8650f76b5d9d5ae71188f3b95d170b54a93.

tony-history-server/test/utils/TestHdfsUtils.java
pdtran3k6: Fixed: https://github.com/linkedin/TonY/pull/81/commits/108fe8650f76b5d9d5ae71188f3b95d170b54a93.

tony-history-server/app/utils/HdfsUtils.java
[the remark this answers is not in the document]
pdtran3k6: So I did some googling. It seems like Java defaults back to what the variable is set to before the exception is thrown. In this case, `intermediateFolders` would be set to an empty `ArrayList`, and that was my assumption. 
I'll add the `e` the `LOG` call. Fixed: https://github.com/linkedin/TonY/pull/81/commits/ac6fbe127274f0b218a87dd44b97ca0c7c112d57.

tony-history-server/app/controllers/JobsMetadataPageController.java
erwa: A few clarifications:

> make Tony AM read from tony-site.xml

TonYClient would read from `tony-site.xml`, apply passed in args, and then write out `tony-final.xml`, which is what the AM uses (no changes in AM).

> write some intermediate script to extract the location from tony-site.xml and inject it into application.conf

`startTHS.sh` can do this and set `-Dprop1=val1` Java system properties that will override what's in `application.conf`. For inspiration, take a look at what Dr. Elephant does here: https://github.com/linkedin/dr-elephant/blob/master/scripts/start.sh

> I don't think that would make the project more maintainable as we're introducing an extra script to maintain, and the location of the job history folder doesn't (and shouldn't) change once we deploy this to prod, so the benefit of having a single source of truth (tony-site.xml) is also very minimal.

Right now, it's just one variable, but in the future, there might be more variables that we need to be consistent between client and server. We should code using best practices from the beginning to save us technical debt in the future.

tony-history-server/app/utils/ParserUtils.java
erwa: That commit still seems to be storing times as Strings rather than longs. I think we should store times as longs. It's a more machine-readable representation. Do the conversion to a human-readable representation later depending on what timezone/datetime format we want to render.

tony-history-server/app/utils/ParserUtils.java
erwa: What I meant was: You can use listStatus with a glob, e.g. `fs.listStatus("<appId>*.jhist")` and then you can check that the returned FileStatus array has length 1.

erwa: For checking the list is exactly size 1, a common pattern is using Guava's Preconditions: `Preconditions.checkArgument(files.size() == 1)`

tony-history-server/app/utils/ParserUtils.java
erwa: What I meant is: I don't think this method should take in a String array. It should take in the full file name and do the parsing/validation internally. Don't you think it's a much cleaner API to just pass a full file name and have a method do validation on it, than for an API to expect the caller to pre-split the filename first?

Please also add some Javadoc.

pdtran3k6:
> There also seem to be some Travis CI compilation issues, e.g.:
> 
> ```
> Compiling 4 Java sources to /home/travis/build/linkedin/TonY/tony-history-server/build/playBinary/testClasses...
> /home/travis/build/linkedin/TonY/tony-history-server/test/utils/TestHdfsUtils.java:84: error: cannot find symbol
>     Assert.assertEquals(new ArrayList<Path>(), HdfsUtils.getMetadataFilePaths(fs, histFolder));
>                                                                                   ^
>   symbol:   variable histFolder
>   location: class TestHdfsUtils
> /home/travis/build/linkedin/TonY/tony-history-server/test/utils/TestHdfsUtils.java:92: error: cannot find symbol
>     assertEquals("", HdfsUtils.getJobId(filePath.toString()));
>     ^
>   symbol:   method assertEquals(String,String)
>   location: class TestHdfsUtils
> /home/travis/build/linkedin/TonY/tony-history-server/test/utils/TestHdfsUtils.java:102: error: cannot find symbol
>     assertTrue(HdfsUtils.isJobFolder(filePath1, regex1));
> ```

Yep, I'm fixing it.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: So right now the UI call getters to retrieve the time. So if we were to store times as longs, we would have to do the date parsing in getters, which looks unusual imo (https://gist.github.com/pdtran3k6/d19e233cf876e8115a7a2db82c7ac32e), but we can certainly do that if that's more maintainable for you guys.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Oh, that's cool, I didn't know `fs.listStatus` could consume regex.

tony-history-server/app/utils/ParserUtils.java
erwa: MR's JobInfo stores times internally as longs. It then provides getXXXTimeStr() methods (in addition to `long getXXXTime()` methods) that return as Strings (in some predefined format). I think we can do something similar.

tony-history-server/app/controllers/JobsMetadataPageController.java
pdtran3k6: Noted. Will work on this in the upcoming PR. @pdtran3k6

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Is it standard to store these as `Date` objects?

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Agreed. Fixed: https://github.com/linkedin/TonY/pull/81/commits/ed06cf126ca618f794531e647344310f689b6969.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Discussed offline, `fs.listStatus` uses `PathFilter` (no glob) which makes the code more complex than it should be. Fixed: https://github.com/linkedin/TonY/pull/81/commits/ed06cf126ca618f794531e647344310f689b6969.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Or I can just convert + format it when UI fetches `started`/`completed`. Also removed some unused methods: https://github.com/linkedin/TonY/pull/81/commits/d150f40c5181d7424751119a86fc9954da1cfd17.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: I agree. But in our case, I think the `HdfsUtils` API is more intuitive and easier to use with the name `getJobId`, compared to a more generic name like `getLastComponent`.

tony-history-server/app/utils/HdfsUtils.java
pdtran3k6: Right. Fixed: https://github.com/linkedin/TonY/pull/81/commits/8b8308d29c6be02cdcafb7a2e32a4a7b0d028316.

tony-history-server/app/utils/HdfsUtils.java
erwa: How about `getJobIdComponent` then?

tony-history-server/app/utils/ParserUtils.java
erwa: Nice, that's better. `Date` is a good choice.

Issue #84, opened 2018-11-16
TonyApplicationMaster doesn't return correct status.


pdtran3k6:
When running locally with a python script that exits 1 (mock a fail worker), TonyAM only fails when `tony.application.single-node` is set to true. We expect that when a non-chief worker fails, TonyAM should continue on training, but it should return fail, regardless of training mode.

PR #86, opened 2018-11-16
1_Tony AM returns FAILED status when a non-chief worker fails

pdtran3k6:
Fix for https://github.com/linkedin/TonY/issues/84.

### DIFF
--- a/tony-history-server/app/controllers/JobsMetadataPageController.java
+++ b/tony-history-server/app/controllers/JobsMetadataPageController.java
@@ -13,7 +13,6 @@ import play.Logger;
 import play.Logger.ALogger;
 import play.mvc.Controller;
 import play.mvc.Result;
-import hadoop.Security;
 import utils.HdfsUtils;
 
 import static utils.HdfsUtils.*;
@@ -21,8 +20,9 @@ import static utils.ParserUtils.*;
 
 
 public class JobsMetadataPageController extends Controller {
-  private static final ALogger LOG = Logger.of(JobsMetadataPageController.class);
+private static final ALogger LOG = Logger.of(JobsMetadataPageController.class);
   private final Config config;
+  private static final String JOB_FOLDER_REGEX = "^application_\\d+_\\d+$";
 
   @Inject
   public JobsMetadataPageController(Config config) {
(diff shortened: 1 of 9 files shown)


pdtran3k6:
In order to test this, we need to make several fundamental changes to `TonyApplicationMaster`. I think we should discuss this offline.

oliverhu:
Talked offline, we should refactor the `monitor()` function a bit.

erwa:
I think this is testable even with our existing code. To trigger the

```
      if (session.isTrainingFinished()) {
        LOG.info("Training has finished.");
        return session.getFinalStatus() == FinalApplicationStatus.SUCCEEDED;
      }	
```

block, create a TonY job with 2 workers and have the Python script exit 1 if it's worker 0 and run forever if it's not worker 0. When worker 0 finishes, `trainingFinished` will be set to `true` and the above code should trigger. Check that `false` is returned.

I think you might need to refactor `TonyApplicationMaster.main()` though. Right now, there's no easy way to inspect the value of `succeeded`. But you could get access to it by refactoring `main()` to:

```
public static void main(...) {
  boolean succeeded = run(args);
    if (succeeded) {
      LOG.info("Application Master completed successfully. exiting");
      System.exit(0);
    } else {
      LOG.info("Application Master failed. exiting");
      System.exit(-1);
    }
}
```

and move the rest of the code into `run()`. Then you should be able to directly call `run()` in your test and inspect the return value.

oliverhu:
@pdtran3k6 it would be easier to read if you have a .diff to indicate what is removed and what is added.

pdtran3k6:
https://gist.github.com/pdtran3k6/0cd3eef0e49958dac1fa778f41b22138

pdtran3k6:
> @pdtran3k6 it would be easier to read if you have a .diff to indicate what is removed and what is added.

https://gist.github.com/pdtran3k6/0cd3eef0e49958dac1fa778f41b22138

erwa:
> https://gist.github.com/pdtran3k6/0cd3eef0e49958dac1fa778f41b22138

Left some comments in there but overall looks good.

pdtran3k6:
Discussed offline with @oliverhu, we think that `runTonyAM` should be refactored to `run` and it shouldn't be static in `TonyApplicationMaster`.

PR #90, opened 2018-11-16
Fix #89 TonY assumes python_venv.zip at root folder

### LINKED ISSUE #89: TonY assumes python_venv.zip at root folder
`TaskExecutor` uses the -python_venv passed by user as the path to locate the python_venv path, this is wrong.. it will always be at root folder and we should make a constant.

oliverhu:
Fix #89


tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -391,13 +392,33 @@ public ContainerLaunchContext createAMContainerSpec(ApplicationId appId, long am
     acls.put(ApplicationAccessType.MODIFY_APP, " ");
     amContainer.setApplicationACLs(acls);
 
+    String command = TonyClient.buildCommand(amMemory, taskParams, pythonBinaryPath, pythonVenv,
+        executes, hdfsClasspath, shellEnv, containerEnv);
+
+    LOG.info("Completed setting up Application Master command " + command);
+    List<String> commands = new ArrayList<>();
+    commands.add(command);
+    amContainer.setCommands(commands);
```
erwa: Might be a little cleaner to just use
```
amContainer.setCommands(ImmutableList.of(command));
```
Saves a couple lines.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -391,13 +392,33 @@ public ContainerLaunchContext createAMContainerSpec(ApplicationId appId, long am
...
+    if (tokens != null) {
+      amContainer.setTokens(tokens);
+    }
+    amContainer.setEnvironment(containerEnv);
+    amContainer.setLocalResources(localResources);
+
+    return amContainer;
+  }
+
+  @SuppressWarnings("WeakerAccess")
```
erwa: Can we remove this and just make the method package-private (no modifier) instead?
```
static String buildCommand(...) { ... }
```

PR #91, opened 2018-11-17
Replace junit import with testNG

oliverhu:
However, when I did a code search, I found all jobhistory tests import junit @erwa  @pdtran3k6 , maybe @pdtran3k6 should also clean up those tests to use testNG.


pdtran3k6:
Ok sure 👍

PR #94, opened 2018-11-19
Add more logging for token fetching logic

oliverhu:
From #74 , if user runs in insecure cluster but didn't disable security check, tony throws exception without enough information.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -35,6 +35,7 @@ import org.apache.commons.cli.ParseException;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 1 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -575,6 +583,7 @@ private ByteBuffer getTokens() throws IOException, URISyntaxException, YarnExcep
                 + "other namenode: " + namenodeUri);
           }
           cred.addToken(otherFSToken.getService(), otherFSToken);
+          LOG.info("Fetched HDFS token for " + nnUri);
```
erwa: Suggest also adding a matching `"Fetching HDFS token for " + nnUri"` LOG statement. This lets us measure token fetching delay.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -546,22 +546,30 @@ private ByteBuffer getTokens() throws IOException, URISyntaxException, YarnExcep
     if (!this.secureMode) {
       return null;
     }
+    LOG.info("Running in secure cluster mode. Fetching delegation token..");
```
hungj: token -> tokens

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
hungj: agreed. Also would be nice to have logging for default FS delegation token as well

PR #99, opened 2018-11-20
LIHADOOP-42483: Publish tony-history-server zip to Artifactory

erwa:
These changes are needed to publish the tony-history-server zip internally to LinkedIn's internal Artifactory instance.


erwa:
I've tested these changes in our internal multiproduct wrapper and the generated artifact-spec.json looks right.

Issue #97, opened 2018-11-20
TonyClient and startTHS.sh should read tony-site.xml config file from TONY_CONF_DIR


erwa:
TonY clients and the TonY History Server need to use the same value for the location of the history files. The client needs to tell the TonY AM to write to that location and the history server needs to read from that location.

We should define this location in a tony-site.xml file and expect it to be in the directory pointed to by the environment variable `TONY_CONF_DIR`, which can be set before running TonY clients or the history server.

See https://github.com/linkedin/TonY/pull/81#discussion_r234392784 for more context.

pdtran3k6:
With that being said, my assumption would be that on both client host and history server host, there would be an environment variable `TONY_CONF_DIR` that points to a `tony-site.xml` with respect to that host. (The client host doesn't need other history config like keytab, principal, etc., but if that increases the maintenance overhead of `tony-site.xml`, then we could make them the same for now).

Issue #96, opened 2018-11-20
TonY uses both tony.history.location and tony.historyFolder


erwa:
Currently, tony-default.xml contains `tony.history.location` whereas other places use `tony.historyFolder`. We should standardize on one. Looking at Hadoop configs, seems like `all.lowercase.period.separated` is the more standard naming convention, rather than using camelCase in config names.

We should also avoid hardcoding `tony.historyFolder` in multiple places and instead define a String constant once and use that elsewhere.

erwa:
Note that `TONY_HISTORY_LOCATION` is already defined in TonyConfigurationKeys.java.

pdtran3k6:
> Note that `TONY_HISTORY_LOCATION` is already defined in TonyConfigurationKeys.java.

For TonyClient, retrieving this value is easy. However, for `tony-history-server`, although I know that there's prolly a way to import `TonyConfigurationKeys.java` and use it within the controllers code, I don't know how, so I'm going to work on https://github.com/linkedin/TonY/issues/97 first, and hardcode the key (tony.history.location + default value) for now (at least inside tony history server code).

PR #102, opened 2018-11-21
2_Tony History Server reads from tony-site.xml

pdtran3k6:
Addressing history server side of https://github.com/linkedin/TonY/issues/97.
Also, acknowledging that certain strings (`tony.history.location`) are still hardcoded. See https://github.com/linkedin/TonY/issues/96#issuecomment-440500639.


tony-history-server/xmlToPlayArgs.py
```diff
@@ -0,0 +1,47 @@
+import xml.etree.ElementTree as ET
+import sys
+
+# This script parses tony-site.xml, then returns a string that
+# includes all the necessary flags for Play to consume.
+# Returns empty string if there's an unrecognizable key in
+# tony-site.xml
+
+if (len(sys.argv) < 2):
+    print("Usage: python xmlToPlayArgs.py <path/to/tony-site.xml>")
```
oliverhu: prefer to use lower case + underscore for python file names

pdtran3k6: 👍

tony-history-server/startTHS.sh
```diff
@@ -181,10 +177,25 @@ save () {
...
     echo " "
 }
-JAVA_OPTS=$(save "$@")
+
+# Set up TONY_CONF_DIR if not exist
+if [[ -z "${TONY_CONF_DIR}" ]]; then
+  echo "TONY_CONF_DIR not set. Default to /export/apps/tony"
+  TONY_CONF_DIR="/export/apps/tony"
+else
+  TONY_CONF_DIR="${TONY_CONF_DIR}"
```
erwa: Is this `else` block necessary? Looks like an identity. Could you just proceed with using `$TONY_CONF_DIR` below without this assignment?

tony-history-server/conf/application.example.conf
```diff
@@ -1,9 +1,10 @@
...
-  location = "/path/to/abc.keytab"
-}
+
 tony {
-  historyFolder = "/path/to/history/folder/"
+  keytab {
+    user = "user1"
+    location = "/some/path/to/tony.keytab"
+  }
+  history.location = "/Users/phattran/.tony/system"
```
erwa: Can we make the example `tony.history.location` more generic, something like `/tmp/tony-history`?

tony-history-server/app/controllers/JobConfigPageController.java
```diff
@@ -37,7 +37,7 @@ public Result index(String jobId) {
     }
 
     List<JobConfig> listOfConfigs;
-    String tonyHistoryFolder = config.getString("tony.historyFolder");
+    String tonyHistoryFolder = config.getString("tony.history.location");
```
erwa: You're planning to replace with `TonyConfigurationKeys.TONY_HISTORY_LOCATION` in a follow-up PR?

pdtran3k6: Yeah, we can discuss this in the sync-up tmr.

tony-history-server/conf/application.example.conf
pdtran3k6: Since this is already merged, I'm going to include this in https://github.com/linkedin/TonY/pull/103.

tony-history-server/startTHS.sh
pdtran3k6: Also including this in https://github.com/linkedin/TonY/pull/103.

PR #103, opened 2018-11-22
3_TonyClient reads history location from tony-site.xml

pdtran3k6:
Addressing https://github.com/linkedin/TonY/issues/97, from the TonyClient side. Also did some minor refactoring.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -73,6 +73,8 @@ import org.apache.hadoop.yarn.server.nodemanager.containermanager.runtime.Contai
 import org.apache.hadoop.yarn.util.ConverterUtils;
 import org.apache.hadoop.yarn.util.Records;
 
+import static com.linkedin.tony.Constants.*;
+
 
 /**
  * User entry point to submit tensorflow job.
@@ -82,9 +84,10 @@ public class TonyClient implements AutoCloseable {
 
   private static final String APP_TYPE = "TENSORFLOW";
   private static final String ARCHIVE_SUFFIX = "tony_archive.zip";
-  private static final String CORE_SITE_CONF = YarnConfiguration.CORE_SITE_CONFIGURATION_FILE;
-  private static final String HADOOP_CONF_DIR = ApplicationConstants.Environment.HADOOP_CONF_DIR.key();
-  private static final String HDFS_SITE_CONF = "hdfs-site.xml";
+  private static final String TONY_CONF_DIR = Constants.TONY_CONF_DIR;
+  private static final String TONY_SITE_CONF = Constants.TONY_SITE_CONF;
+  private static final String TONY_DEFAULT_XML = Constants.TONY_DEFAULT_XML;
+  private static final String TONY_XML = Constants.TONY_XML;
 
   // Configurations
   private YarnClient yarnClient;
@@ -142,7 +145,6 @@ public class TonyClient implements AutoCloseable {
   private boolean run() throws IOException, InterruptedException, URISyntaxException, YarnException {
     LOG.info("Starting client..");
     yarnClient.start();
-
     YarnClientApplication app = yarnClient.createApplication();
     GetNewApplicationResponse appResponse = app.getNewApplicationResponse();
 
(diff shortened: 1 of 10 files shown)


tony-core/src/main/java/com/linkedin/tony/Utils.java
```diff
@@ -416,5 +421,24 @@ public static void printWorkerTasksCompleted(AtomicInteger completedWTasks, long
     LOG.info("Completed worker tasks: " + completedWTasks.get() + " out of " + totalWTasks + " worker tasks." );
   }
 
+  public static Configuration addResourceToConfig(Configuration tonyConf, CommandLine cliParser) {
```
oliverhu: this refactoring looks weird. it is fine to keep this portion of code inside `init()` function.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -82,9 +82,13 @@
 
   private static final String APP_TYPE = "TENSORFLOW";
   private static final String ARCHIVE_SUFFIX = "tony_archive.zip";
-  private static final String CORE_SITE_CONF = YarnConfiguration.CORE_SITE_CONFIGURATION_FILE;
-  private static final String HADOOP_CONF_DIR = ApplicationConstants.Environment.HADOOP_CONF_DIR.key();
-  private static final String HDFS_SITE_CONF = "hdfs-site.xml";
+  private static final String CORE_SITE_CONF = Constants.CORE_SITE_CONF;
+  private static final String HADOOP_CONF_DIR = Constants.HADOOP_CONF_DIR;
+  private static final String HDFS_SITE_CONF = Constants.HDFS_SITE_CONF;
+  private static final String TONY_CONF_DIR = Constants.TONY_CONF_DIR;
```
erwa: No need to redefine all these constants here if they're already defined in `Constants`. Just directly reference `Constants.XXX` when needed.

tony-core/src/main/java/com/linkedin/tony/Utils.java
erwa: Agreed. I think this could be a private method in `TonyClient` called `initTonyConf()`. What the method does seems specific to the TonyClient (preparing its Configuration) rather than a general utility method.

tony-core/src/test/java/com/linkedin/tony/TestTonyConfigurationFields.java
```diff
@@ -32,6 +32,9 @@ public void initializeMemberVariables() {
     xmlPropsToSkipCompare.add(TonyConfigurationKeys.getGPUsKey(Constants.WORKER_JOB_NAME));
     xmlPropsToSkipCompare.add(TonyConfigurationKeys.getResourcesKey(Constants.WORKER_JOB_NAME));
     configurationPropsToSkipCompare.add(TonyConfigurationKeys.DOCKER_IMAGE);
+    // These properties are in TonyConfigurationKeys, but not in tony-default.xml
+    // so we skip this check
+    configurationPropsToSkipCompare.add(TonyConfigurationKeys.TONY_HISTORY_LOCATION);
```
erwa: Why do you want to skip checking this config? I think it's informative to have a default value present in `tony-default.xml`.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
oliverhu: +1

tony-core/src/main/java/com/linkedin/tony/Utils.java
pdtran3k6: 👍

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
pdtran3k6: 👍

tony-core/src/test/java/com/linkedin/tony/TestTonyConfigurationFields.java
pdtran3k6: Discussed offline. We should not skip this check as all the configurations in tony-site.xml will also be in tony-default.xml.

PR #107, opened 2018-11-26
6_WIP_TonyAM_dispatchEvents

pdtran3k6:


tony-core/build.gradle
```diff
@@ -13,6 +16,7 @@ dependencies {
   compile deps.hadoop.yarn_client
   compile deps.hadoop.yarn_common
   compile deps.external.zip4j
+  compile deps.external.avro
```
erwa: Let's keep alphabetized

tony-core/build.gradle
```diff
@@ -1,3 +1,6 @@
+plugins {
+  id "com.commercehub.gradle.plugin.avro" version "0.9.0"
```
erwa: Just by applying this plugin, do the Avro classes automatically get generated as part of build?

tony-core/src/main/java/com/linkedin/tony/events/HistoryEvent.java
```diff
@@ -0,0 +1,26 @@
...
+import org.apache.hadoop.classification.InterfaceAudience;
+import org.apache.hadoop.classification.InterfaceStability;
+
+/**
+ * Interface for event wrapper classes.  Implementations each wrap an
+ * Avro-generated class, adding constructors and accessor methods.
+ */
+@InterfaceAudience.Private
+@InterfaceStability.Unstable
+public interface HistoryEvent {
```
erwa: Let's also add a `long getTimestamp()` method

tony-core/build.gradle
pdtran3k6: Yes.

tony-core/src/main/java/com/linkedin/tony/events/HistoryEvent.java
pdtran3k6: I'm going to include this in the next PR where I work on the reading part of the history files.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -0,0 +1,108 @@
...
+      eventQueue.put(event);
+    } catch (InterruptedException e) {
+      LOG.error("Failed to add event " + event + " to event queue", e);
+    }
+  }
+
+  private void writeEvent(BlockingQueue<Event> queue, DataFileWriter<Event> writer) {
+    Event event = null;
+    try {
+      event = queue.poll();
```
erwa: I think you want to use `take()` so that this will block if the queue is empty.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -0,0 +1,108 @@
...
+
+  public void emitEvent(Event event) {
+    try {
+      eventQueue.put(event);
+    } catch (InterruptedException e) {
+      LOG.error("Failed to add event " + event + " to event queue", e);
+    }
+  }
+
+  private void writeEvent(BlockingQueue<Event> queue, DataFileWriter<Event> writer) {
```
erwa: I think you don't need to pass `queue` and `writer` into this method since they are instance variables already.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -0,0 +1,108 @@
...
+      out.close();
+    } catch (IOException e) {
+      LOG.error("Failed to close writer", e);
+    }
+
+    // At this point, historyFile should be set
+    // If not, then discard all events
+    if (historyFile == null) {
+      try {
+        myFs.delete(tmpFile, true);
```
erwa: Let's log a warning message before we discard all events.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.avro.io.DatumWriter;
+import org.apache.avro.specific.SpecificDatumWriter;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.FileUtil;
+import org.apache.hadoop.fs.Path;
+
+
+public class EventHandler implements Runnable {
```
erwa: Please add some tests for this EventHandler

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
pdtran3k6: Yep I tried that. The problem is that because it's blocking, the thread would hang forever, and there's no obvious way to let the queue know that it should proceed. We can discuss more offline if you wanted to.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
pdtran3k6: I just think it would be easier to write test this way.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
pdtran3k6: Yep.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
erwa: You should use Thread.interrupt() in your `stop()` method and catch an `InterruptedException` here.

erwa:
This branch has some conflicts that need to be resolved manually.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
pdtran3k6: Added.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
pdtran3k6: Discussed offline, we will go with a different approach (emitting EOQ event).

pdtran3k6:
For https://github.com/linkedin/TonY/issues/129, we can't really move ParserUtils to tony-core due to other dependencies related to view rendering (Scala template to HTML), so for now I suggest we should just have a separate method and merge the PR.
Eventually, when we switch to a more robust UI library, we can then refactor the code to remove duplicate method.

erwa:
> For #129, we can't really move ParserUtils to tony-core due to other dependencies related to view rendering (Scala template to HTML)

Can you be more specific on what the issue is? Scala code should be able to convert a `long` date to a `String` date in any date format you like.

pdtran3k6:
> > For #129, we can't really move ParserUtils to tony-core due to other dependencies related to view rendering (Scala template to HTML)
> 
> Can you be more specific on what the issue is? Scala code should be able to convert a `long` date to a `String` date in any date format you like.

Resolved offline.

erwa:
```
    // Clear the queue
    while (!eventQueue.isEmpty()) {
      drainQueue(eventQueue, dataFileWriter);
    }
```

`drainQueue` sounds like it should include the `while` loop inside it. Move the `while` inside the `drainQueue` method?

erwa:
```
private Path tmpFile = new Path(Constants.TMP_AVRO);
```

Why do you write to a temp file first and then later move it? I think you can just directly write to the final output location. The TonY History server can just read as many records as it can from the final file and ignore partially written files. (This is what the MR history server does.)

Also, I think you still need to implement the THS logic that moves files from the `intermediate` directory to the `finished` directory.

pdtran3k6:
> ```
> private Path tmpFile = new Path(Constants.TMP_AVRO);
> ```
> Why do you write to a temp file first and then later move it? Why can't you just directly write to the final output location?

Initially you don't have the job's status and completed time so you can't generate jhist file name.

pdtran3k6:
> ```
>     // Clear the queue
>     while (!eventQueue.isEmpty()) {
>       drainQueue(eventQueue, dataFileWriter);
>     }
> ```
> `drainQueue` sounds like it should include the `while` loop inside it. Move the `while` inside the `drainQueue` method?

Sure.

erwa:
> > ```
> > private Path tmpFile = new Path(Constants.TMP_AVRO);
> > ```
> > Why do you write to a temp file first and then later move it? Why can't you just directly write to the final output location?
> 
> Initially you don't have the job's status and completed time so you can't generate jhist file name.

Gotcha. I had an idea, which we can do in a follow-up PR: We can write the temporary file to the same folder as the final file but with a name like `<appId>-<startTime>-INPROGRESS-<user>.jhist`. The THS can read the temporary file and show in-progress apps as well.

Currently, it seems the tmpFile is written to the current working directory (or maybe without a full path, it defaults to the user's home directory). Can we write it to the same directory as the final output location?

tony-core/src/test/resources/nestedHistFolder/2017/07/job3/metadata.json
```diff
@@ -0,0 +1,2 @@
+// Copyright 2018 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
+// See LICENSE in the project root for license information.
```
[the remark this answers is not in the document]
pdtran3k6: No we don't since I already excluded .json files in build.gradle

tony-core/src/test/java/com/linkedin/tony/TestHdfsUtils.java
```diff
@@ -84,32 +89,32 @@ public void testPathExists_throwsException() throws IOException {
...
-    assertEquals("", HdfsUtils.contentOfHdfsFile(fs, filePath));
+    assertEquals(HdfsUtils.contentOfHdfsFile(fs, filePath), "");
   }
 
   @Test
   public void testGetJobId_typicalCase() {
-    Path filePath1 = new Path("./test/resources/job1");
-    Path filePath2 = new Path("./test/resources/app2/");
+    Path filePath1 = new Path("./tony-core/src/test/resources/job1");
+    Path filePath2 = new Path("./tony-core/src/test/resources/app2/");
```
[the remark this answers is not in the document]
pdtran3k6: Sure. 👍

PR #108, opened 2018-11-27
LIHADOOP-41388: Inject Azkaban metadata to TonY configuration

erwa:
Tested on our test cluster. Verified that `/system/tony-history/application_XXX_XXX/config.xml` now contains Azkaban metadata such as

* azkaban.link.execution.url
* azkaban.link.job.url
* azkaban.flow.projectname

etc.

### DIFF
--- a/tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
+++ b/tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
@@ -4,10 +4,8 @@
  */
 package com.linkedin.tony.azkaban;
 
-import azkaban.flow.CommonJobProperties;
 import azkaban.jobtype.HadoopJavaJob;
 import azkaban.jobtype.HadoopJobUtils;
-import azkaban.security.commons.HadoopSecurityManager;
 import azkaban.utils.Props;
 import java.io.File;
 import java.io.FileOutputStream;
@@ -17,8 +15,6 @@ import java.util.Map;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.log4j.Logger;
 
-import static org.apache.hadoop.security.UserGroupInformation.HADOOP_TOKEN_FILE_LOCATION;
-
 
 /**
  * The Azkaban jobtype for running a TensorFlow job.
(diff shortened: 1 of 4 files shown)


erwa:
There was a lot of duplicate code in TensorFlowJob type which I removed (HadoopJavaJob (which TensorFlowJob extends) already does most of it).

The code changes to inject the Azkaban metadata were minimal because `TensorFlowJob.run()` calls `super.run()` (`HadoopJavaJob.run()`), which already calls `HadoopConfigurationInjector.prepareResourcesToInject()`.

On the `TonyClient` side, just needed to add `HadoopConfigurationInjector.injectResources(new Props() /* ignored */);` to the `main()` method before a `new Configuration()` is created.

tony-cli/build.gradle
```diff
@@ -14,6 +14,7 @@ shadowJar {
     exclude(dependency('com.sun.jersey.contribs:'))
     exclude(dependency('org.apache.hadoop:'))
   }
+  zip64 true
```
oliverhu: what does this do?

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
```diff
@@ -104,30 +71,13 @@ public String getJavaClass() {
...
-    String typeUserGlobalJVMArgs = getJobProps().getString(HadoopJobUtils.JOBTYPE_GLOBAL_JVM_ARGS, null);
-    if (typeUserGlobalJVMArgs != null) {
-      args += " " + typeUserGlobalJVMArgs;
-    }
-    String typeSysGlobalJVMArgs = getSysProps().getString(HadoopJobUtils.JOBTYPE_GLOBAL_JVM_ARGS, null);
-    if (typeSysGlobalJVMArgs != null) {
-      args += " " + typeSysGlobalJVMArgs;
-    }
-    String typeUserJVMArgs = getJobProps().getString(HadoopJobUtils.JOBTYPE_JVM_ARGS, null);
-    if (typeUserJVMArgs != null) {
```
oliverhu: i remember we left some of these here for compatibility, make sure we don't need to force user to upgrade tony

tony-cli/build.gradle
erwa: It's needed once the number of files in the jar exceeds 65535 entries: https://github.com/johnrengelman/shadow/issues/107

tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TensorFlowJob.java
erwa: This is the exact same code as in HadoopJavaJob.getJVMArguments (which is already called via super.getJVMArguments() on line 72), so it should be safe to remove.

I also tested the jobtype changes in Azkaban (replaced the jar and reloaded the jobtype plugins) and it still worked fine.

tony-cli/build.gradle
erwa: I guess adding the Azkaban dependencies in tony-core tipped the fat jar over 65535 entries.

Issue #112, opened 2018-11-29
TonY should inject the version info (build version, commit, user, branch, etc.) into the Configuration


erwa:
TonY should inject build metadata (version number, commit hash, user, date, etc.) into the Configuration created by the TonyClient so that this info is available in the config.xml for TonY applications.

PR #113, opened 2018-11-29
LIHADOOP-41391: Inject TonY version into TonY configuration

erwa:
Fixes https://github.com/linkedin/TonY/issues/112.

Adapted VersionInfo.java from Hadoop and VersionInfo.gradle (renamed to version-info.gradle) from Dali.

Also moved the XXXUtils files into a util package.

The contents of the `version-info.properties` added to the tony jars look like

```
#Thu, 29 Nov 2018 09:02:46 -0800

version=0.1.5
revision=1cba10da369b846c53a3f99e9acbf8321db876e7
branch=inject-version
user=ahsu
date=Thu Nov 29 09\:02\:46 PST 2018
url=git@github.com\:erwa/TonY.git
srcChecksum=58e8dc8ad20e1c0c827ef429c6e4c567
```

The `tony.version-info.XXX` properties added to the `config.xml` written out by TonY look like:

```
<property><name>tony.version-info.url</name><value>git@github.com:erwa/TonY.git</value><final>false</final><source>programatically</source><source>tony-final.xml</source></property>

<property><name>tony.version-info.branch</name><value>inject-version</value><final>false</final><source>programatically</source><source>tony-final.xml</source></property>

<property><name>tony.version-info.checksum</name><value>58e8dc8ad20e1c0c827ef429c6e4c567</value><final>false</final><source>programatically</source><source>tony-final.xml</source></property>

<property><name>tony.version-info.version</name><value>0.1.5</value><final>false</final><source>programatically</source><source>tony-final.xml</source></property>

<property><name>tony.version-info.date</name><value>Thu Nov 29 09:02:46 PST 2018</value><final>false</final><source>programatically</source><source>tony-final.xml</source></property>

<property><name>tony.version-info.user</name><value>ahsu</value><final>false</final><source>programatically</source><source>tony-final.xml</source></property>

<property><name>tony.version-info.revision</name><value>1cba10da369b846c53a3f99e9acbf8321db876e7</value><final>false</final><source>programatically</source><source>tony-final.xml</source></property>
```


erwa:
Fixed the TestTonyConfigurationFields test.

.gitignore
```diff
@@ -23,6 +23,7 @@
 .settings
 .svn
 .vscode/
+/test/
```
oliverhu: this looks dangerous, would this possibly exclude future test files from git?

gradle/version-info.gradle
```diff
@@ -0,0 +1,78 @@
+import java.security.MessageDigest
```
oliverhu: can you add some documentation what this gradle file does?

tony-core/src/main/java/com/linkedin/tony/util/VersionInfo.java
```diff
@@ -0,0 +1,142 @@
...
+public class VersionInfo {
+  private static final Log LOG = LogFactory.getLog(VersionInfo.class);
+
+  private Properties versionInfo;
+
+  private VersionInfo() {
+    versionInfo = new Properties();
+    String versionInfoFile = "version-info.properties";
+    try {
+      InputStream is = Thread.currentThread().getContextClassLoader()
```
oliverhu: why throw exception and then catch exception and print a log? Can we just if (is == null ) {log; return;}?

.gitignore
erwa: This should only exclude the root `/test` folder since I prefixed with a `/`. Let me test this out.

erwa: Confirmed that only the root `/test` folder is impacted. But I see how this can be confusing and I think it's better that I just ignore/deal with untracked files some other way. (I could just create my test directory in `/tmp/test`, for example.) I'll remove this.

tony-core/src/main/java/com/linkedin/tony/util/VersionInfo.java
erwa: I think it was written this way to avoid duplicate log lines. The current way it's written, there's just one log line on line 38.

gradle/version-info.gradle
erwa: Will add.

erwa:
Addressed @oliverhu's comments in https://github.com/linkedin/TonY/pull/113/commits/c9a6919fd4a7d7fed04bd1e2ae1266cc733a334f.

tony-core/src/main/java/com/linkedin/tony/util/VersionInfo.java
oliverhu: oh i see.. I feel it is more intuitive to just log instead of throwing an exception. Also we can use try-with-resources

```
try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(versionInfoFile)){
if (is == null) {
   LOG
} else {
 versionInfo.load(is);
} catch () {
}
```

Also it looks phishing that getResourceAsStream would return null & throw IOException both. I'd suppose that function should either throw exception or return null?

oliverhu:
Added another comment

tony-core/src/main/java/com/linkedin/tony/util/VersionInfo.java
erwa: Updated to use try-with-resources and just log instead of throw exception if `is` is `null`.

erwa:
Rebased on `master`. Will merge if tests pass.

PR #117, opened 2018-12-03
7_THS_displayEvents

pdtran3k6:


erwa:
Hey, it's difficult to see what the incremental changes in your PR are, since they include all the changes from previous (including unmerged) PRs.

Can you push all your branches to linkedin/TonY and update the base branch of each PR so that each PR only shows the incremental change?

Ref: https://blog.github.com/2016-08-15-change-the-base-branch-of-a-pull-request/

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -368,12 +380,34 @@ private boolean run(String[] args) {
...
+    eventQueue.add(finishEvent);
     // By this time jobDir should have been set
     HistoryFileUtils.createHistoryFile(fs,
-        TonyJobMetadata.newInstance(yarnConf, appIdString, started, completed, succeeded), jobDir);
+        TonyJobMetadata.newInstance(yarnConf, appIdString, started, completed, succeeded), jobDir, eventQueue,
+        writeRetryCnt);
     return succeeded;
   }
 
+  private Event constructEvent(String type) {
```
erwa: I think event construction should actually be done separately for each event, since each event may require custom parameters. E.g.: TASK_INITED would need the task id but APPLICATION_INITED does not.

MRAppMaster just directly does

```
Event event = new XEvent(...);
dispatcher.getEventHandler().handler(event);
```

pdtran3k6:
Ok, I confirmed that I have permission to push branches to linkedin/tony. I will clone and work on the linkedin/tony repo for the next PR. For now, I think you can check the changes of this PR specifically by picking the last commit.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
```diff
@@ -15,13 +21,45 @@
...
+      while (count > 0) {
+        try {
+          dataFileWriter.append(event);
+        } catch (IOException e) {
+          LOG.error("Failed to append event " + event.toString());
+          LOG.error("Retrying...");
+          count--;
+          continue;
+        }
+        count = numRetries;
```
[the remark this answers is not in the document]
pdtran3k6: Line 36 is for initialization. It's not a part of the loop. If an event takes more than 1 attempt to be written to file, the next event should have their number of attempts available reset to `numRetries` again. 
Also I think the naming could improve. `count` should be `attpsLeft`, and `numRetries` could be `maxRetry`. Something like this:
```
...
int attpsLeft = maxRetry;
    Event event;
    while (!eventQueue.isEmpty()) {
      event = eventQueue.poll();
      // Retry writing event until successful
      while (count > 0) {
        try {
          dataFileWriter.append(event);
        } catch (IOException e) {
          LOG.error("Failed to append event " + event.toString());
          LOG.error("Retrying...");
          attpsLeft--;
          continue;
        }
        attpsLeft = maxRetry;
        break;
      }
    }
...
```

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
```diff
@@ -15,13 +21,45 @@
...
+        break;
+      }
+    }
+
+    // We have to close writer before closing OutputStream
+    try {
+      dataFileWriter.close();
+      if (out != null) {
+        out.close();
```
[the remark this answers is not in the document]
pdtran3k6: I think I tried that at some point with not a lot of success as IDEA was saying something along the line of `Can't use a reference within resources block`. It seems like that's not an issue now, 🤔 I'll try again and let you know.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
```diff
@@ -15,13 +21,45 @@
...
     } catch (IOException e) {
       LOG.error("Failed to create history file", e);
     }
+
+    int count = numRetries;
+    Event event;
+    while (eventQueue.peek() != null) {
+      event = eventQueue.poll();
+      // Retry writing event until successful
+      while (count > 0) {
```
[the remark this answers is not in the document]
pdtran3k6: Sure.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
```diff
@@ -15,13 +21,45 @@
...
-      fs.create(historyFile);
+      out = fs.create(historyFile);
+      dataFileWriter.create(Event.SCHEMA$, out);
     } catch (IOException e) {
       LOG.error("Failed to create history file", e);
     }
+
+    int count = numRetries;
+    Event event;
+    while (eventQueue.peek() != null) {
```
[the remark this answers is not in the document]
pdtran3k6: Yep, I could.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
```diff
@@ -15,13 +21,45 @@
...
+      event = eventQueue.poll();
+      // Retry writing event until successful
+      while (count > 0) {
+        try {
+          dataFileWriter.append(event);
+        } catch (IOException e) {
+          LOG.error("Failed to append event " + event.toString());
+          LOG.error("Retrying...");
+          count--;
```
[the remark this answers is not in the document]
pdtran3k6: Erik said HDFS FileSystem does have built-in retries indeed, but I think this retry here is specifically to handle failure when appending events, not for writing to outstream.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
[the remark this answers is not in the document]
pdtran3k6: All the changes that are related to writing events are pushed to https://github.com/linkedin/TonY/pull/107.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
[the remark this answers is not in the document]
pdtran3k6: https://github.com/linkedin/TonY/pull/117#discussion_r238847850.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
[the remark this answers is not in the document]
pdtran3k6: https://github.com/linkedin/TonY/pull/117#discussion_r238847850.

tony-core/src/main/resources/tony-default.xml
```diff
@@ -284,4 +285,12 @@
     <name>tony.init.module</name>
     <value>Startup</value>
   </property>
+
+  <property>
+    <description>
+      Number of times the AM tries to write the event to the history file.
+    </description>
+    <name>tony.history.write.retry-count</name>
```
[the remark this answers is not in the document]
pdtran3k6: or even `tony.history.maxAppends`. And of course, I'll update the description as well.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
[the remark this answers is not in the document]
pdtran3k6: https://github.com/linkedin/TonY/pull/117#discussion_r238847850.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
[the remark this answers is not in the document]
erwa: Got it, didn't notice the nested while loops before.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -93,7 +99,9 @@
   private String amHostPort;
   private static FileSystem fs;
   private static String tonyHistoryFolder;
+  private static int writeRetryCnt;
   private static Path jobDir;
+  private Queue<Event> eventQueue = new LinkedList<>();
```
edwinalu: As Anthony mentioned earlier, this should be a BlockingQueue, and a separate thread can write the events to the history file while the application is running. Since the events are written concurrently, it won't take so much time at the end to write the history file if there are a lot of events. Also, in case of failure, there would be some information in the log, which will be useful for debugging the failure.

The history writing logic can be put in a separate class (HistoryWriter, EventLogger), and the queue could be moved there. It would be started and stopped by the ApplicationMaster, and have methods for enqueuing events.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
```diff
@@ -15,13 +21,45 @@
 public class HistoryFileUtils {
   private static final Log LOG = LogFactory.getLog(HistoryFileUtils.class);
 
-  public static void createHistoryFile(FileSystem fs, TonyJobMetadata metadata, Path jobDir) {
+  public static void createHistoryFile(FileSystem fs, TonyJobMetadata metadata, Path jobDir, Queue<Event> eventQueue, int numRetries) {
```
edwinalu: Instead of static utils, this could be a HistoryWriter class. Please add comments for the class and methods.

tony-history-server/app/controllers/JobEventPageController.java
```diff
@@ -0,0 +1,60 @@
...
+import play.Logger.ALogger;
+import play.mvc.Controller;
+import play.mvc.Result;
+import utils.HdfsUtils;
+
+import static utils.HdfsUtils.*;
+import static utils.ParserUtils.*;
+
+
+public class JobEventPageController extends Controller {
```
edwinalu: Please add comments for the class and methods.

tony-core/src/main/avro/Metric.avsc
```diff
@@ -0,0 +1,9 @@
+{
+  "namespace": "com.linkedin.tony.events",
+  "type": "record",
+  "name": "Metric",
+  "fields": [
```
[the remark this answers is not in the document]
erwa: ApplicationFinished (see above) does store an array of Metric.

tony-history-server/app/controllers/JobEventPageController.java
```diff
@@ -0,0 +1,60 @@
...
+  private static final ALogger LOG = Logger.of(JobEventPageController.class);
```
erwa: ALogger is a funny class name. I wonder why Play named it that as opposed to just Logger.

tony-history-server/app/cache/CacheWrapper.java
```diff
@@ -24,9 +25,18 @@
...
+   * - key: job ID (application_[0-9]+_[0-9]+)
+   * - value: List of JobEvent objects. Each JobEvent object
+   * represents an Event in job's jhist
+   */
+  private static Cache<String, List<JobEvent>> eventCache;
+
   public CacheWrapper() {
     metadataCache = CacheBuilder.newBuilder().build();
     configCache = CacheBuilder.newBuilder().build();
+    eventCache = CacheBuilder.newBuilder().build();
```
erwa: Do these caches have a max size? Should we set a max size? What is their behavior when memory is scarce?

tony-history-server/app/controllers/JobEventPageController.java
```diff
@@ -0,0 +1,60 @@
...
+      return internalServerError("Failed to initialize file system");
+    }
+
+    List<JobEvent> listOfEvents;
+    Path tonyHistoryFolder = new Path(config.getString(TonyConfigurationKeys.TONY_HISTORY_LOCATION));
+    listOfEvents = cache.getIfPresent(jobId);
+    if (listOfEvents != null) {
+      return ok(views.html.event.render(listOfEvents));
+    }
+    List<Path> jobFolder = getJobFolders(myFs, tonyHistoryFolder, jobId);
```
erwa: You could reduce the number of `return` statements by restructuring as follows:

```
listOfEvents = cache.getIfPresent(jobId);
if (listOfEvents == null) {
  // read from file

  if (listOfEvents.size() == 0) {
    ...
  }
  cache.put(...);
}

return ok(...);
```

tony-history-server/app/utils/ParserUtils.java
```diff
@@ -144,5 +160,55 @@ public static JobMetadata parseMetadata(FileSystem fs, Path jobFolderPath, Strin
...
+    } catch (IOException e) {
+      LOG.error("Failed to open history file", e);
+    }
+
+    DatumReader<Event> datumReader = new SpecificDatumReader<>(Event.class);
+    List<JobEvent> events = new ArrayList<>();
+    try (DataFileStream<Event> avroFileStream = new DataFileStream<>(in, datumReader)) {
+      Event record;
+      while (avroFileStream.hasNext()) {
+        record = avroFileStream.next();
```
erwa: I think you can use the `.next(T) ` method call to reuse the existing `Event` object: https://avro.apache.org/docs/1.7.7/api/java/org/apache/avro/file/DataFileStream.html#next(D)

I think you'll to initialize to a dummy `Event` on line 191, though.

tony-history-server/app/controllers/JobEventPageController.java
pdtran3k6: I guess this is a style thing. I prefer having more return statements to make the code flatter rather than nested code to reduce # of returns.

tony-history-server/app/utils/ParserUtils.java
pdtran3k6: Yeah I saw that too but I thought that was kinda weird so I went with this instead.

tony-history-server/app/controllers/JobEventPageController.java
erwa: True, this is mainly stylistic. One argument in favor of my approach is less code duplication. In your approach, you duplicate the line `return ok(views.html.event.render(listOfEvents));`. I try to follow the DRY (Don't Repeat Yourself) principle.

tony-history-server/app/utils/ParserUtils.java
erwa: We'd have to benchmark it to be certain, but I think reusing the same `Event` object would be more performant, since it saves a lot of unnecessary object allocations and garbage collection.

pdtran3k6: Fixed.

tony-core/src/main/java/com/linkedin/tony/util/HistoryFileUtils.java
pdtran3k6: I put the logic in `EventHandler` class now. See https://github.com/linkedin/TonY/pull/117/files#diff-f54d0d666f357be10268a94d9a9384a7 for more details.

tony-history-server/app/cache/CacheWrapper.java
erwa: Please specify a maximumSize when creating these caches. (This should be configurable in the tony-site.xml file.) Otherwise, we'll eventually run out of memory and the service will crash.

erwa:
This branch also has conflicts that need manual resolution.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -0,0 +1,112 @@
...
+    } catch (InterruptedException e) {
+      LOG.error("Failed to add event " + event + " to event queue", e);
+    }
+  }
+
+  @VisibleForTesting
+  public void writeEvent(BlockingQueue<Event> queue, DataFileWriter<Event> writer) {
+    Event event = null;
+    try {
+      event = queue.poll();
```
erwa: This should be switched to .take() and we should use interrupt to stop the event handler.

Currently, we are doing a lot of spinning:
```
while (!isStopped) {
  writeEvent(...);
}
``` 

is going to use a lot of CPU looping over and over again.

erwa: Filed https://github.com/linkedin/TonY/issues/128 for this.

tony-core/src/test/java/com/linkedin/tony/TestEventHandler.java
```diff
@@ -0,0 +1,155 @@
...
+  private static final Log LOG = LogFactory.getLog(TestEventHandler.class);
+  private FileSystem fs = null;
+  private BlockingQueue<Event> eventQueue;
+  private Thread eventHandlerThread;
+  private Event eEventWrapper;
+  private ApplicationInited eAppInitEvent = new ApplicationInited("app123", 1, "fakehost");
+  private Path jobDir = new Path("./src/test/resources/jobDir");
+  private TonyJobMetadata metadata = new TonyJobMetadata();
+
+  private List<Event> parseEvents(FileSystem fs, Path jobDir, TonyJobMetadata metadata) {
```
erwa: Please don't duplicate the code in `ParserUtils.parseEvents()` here. You can move `ParserUtils` from the tony-history-server module to the tony-core module to avoid the cyclic dependency.

PR #118, opened 2018-12-03
1_Updated docs

pdtran3k6:


tony-history-server/README.md
```diff
@@ -1,12 +1,15 @@
...
 ```
 $ cd tony-history-server
 ```
 
+- Rename `conf/application.example.conf` to `conf/application.conf` and set 
+all the necessary configurations.
+
 - Build the project with Gradle (installed):
 ```
 $ gradle build
```
erwa: Can we use `./gradlew`?

tony-history-server/README.md
```diff
@@ -35,21 +38,29 @@ $ gradle testPlayBinary
...
 
-## Deploy to test cluster
-Before using the script, ensure that you have set execution permission (`chmod +x buildAndDeploy.sh`)
-
-- To bundle the history server app and deploy to another host
+## Production
+- Bundle the production binary:
 ```
-$ ./buildAndDeploy.sh user hostname.test.abc.com
+$ gradle dist
```
erwa: Same

tony-history-server/conf/tony-site.sample.xml
```diff
@@ -0,0 +1,47 @@
...
+        <name>tony.https.keystore.algorithm</name>
+        <value>SunX509</value>
+    </property>
+    <property>
+        <name>tony.http.port</name>
+        <value>disabled</value>
+    </property>
+    <property>
+        <name>tony.secret.key</name>
+        <value>Prod</value>
```
erwa: What does this `secret.key` do? The keystore password makes sense to me, but I'm not sure what this `secret.key` is for. Is the `secret.key` a password of sorts or is it just an identifier/label?

tony-history-server/README.md
pdtran3k6: Sure.

tony-history-server/conf/tony-site.sample.xml
pdtran3k6: https://www.playframework.com/documentation/2.6.x/ApplicationSecret.

tony-history-server/README.md
pdtran3k6: Sure.

tony-history-server/conf/tony-site.sample.xml
erwa: Thanks. That doc says "Anyone that can get access to the secret will be able to generate any session they please, effectively allowing them to log in to your system as any user they please."

So we should make sure the production secret we use is not checked in to source control and not in the tony-site.xml on the client side.

erwa: I would suggest changing the value of `tony.secret.key` in the example to `changeme` (which is the default for `play.http.secret.key`.

Issue #123, opened 2018-12-11
TonY History Server should start using nohup


erwa:
Currently, the startTHS.sh script uses `exec` to start, and so if the SSH session in which the THS is started dies, the THS process itself dies. We should use `nohup` so the THS will continue running even when the SSH session is closed. (Alternatively, we could run THS as a background process.)

PR #125, opened 2018-12-11
1_Updated + added new script

pdtran3k6:
For https://github.com/linkedin/TonY/issues/123.

### DIFF
--- a/tony-history-server/stopTHS.sh
+++ b/tony-history-server/stopTHS.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env sh
+###########################################################################################################
+# NAME: stopTHS.sh
+#
+# DESCRIPTION:
+# This script stops the Tony History Server (THS) process.
+#
+#
+# INPUT:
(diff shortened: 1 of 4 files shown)


tony-history-server/stopTHS.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# EXIT CODE:
+# 0 - Success
+# 1 - Failed to find RUNNING_PID
+#
+#
+# CHANGELOG:
+# DEC 10 2018 PHAT TRAN
+############################################################################################################
+RUNNING_PID_PATH=./RUNNING_PID
```
erwa: Currently, it seems like startTHS.sh does not create this file. Should startTHS.sh be updated?

erwa: Synced offline. Apparently, `nohup` automatically generates a `RUNNING_PID` file.

Would be good to add a comment to startTHS.sh saying that nohup will generate the RUNNING_PID file.

Issue #124, opened 2018-12-11
TonyClient should print a link to the TonY History Server page for an application when it finishes


erwa:
To make the THS page for a TonY application more easily accessible, the TonyClient could print a link to it when the application finishes.

PR #126, opened 2018-12-11
1_Added app THS link printer

pdtran3k6:
For https://github.com/linkedin/TonY/issues/124.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/Constants.java
+++ b/tony-core/src/main/java/com/linkedin/tony/Constants.java
@@ -13,6 +13,11 @@ public class Constants {
   // File Permission
   public static final FsPermission perm770 = new FsPermission((short) 0770);
 
+  // History Server related constants
+  public static final String JH_HOST= "ltx1-mlearn-alpha-jh01.grid.linkedin.com";
+  public static final String JOBS_SUFFIX = "jobs";
+  public static final String CONFIG_SUFFIX = "config";
+
   // TensorFlow constants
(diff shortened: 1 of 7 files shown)


tony-core/src/main/java/com/linkedin/tony/Constants.java
```diff
@@ -13,6 +13,11 @@
   // File Permission
   public static final FsPermission perm770 = new FsPermission((short) 0770);
 
+  // History Server related constants
+  public static final String JH_HOST= "ltx1-mlearn-alpha-jh01.grid.linkedin.com";
```
oliverhu: don't put this....

pdtran3k6: Hmm ok yeah I was thinking that we have our own private network so they can't even access it from outside, but it's good practice anw. I'll update this and other instances that uses real hostname/path/etc.

tony-core/src/main/java/com/linkedin/tony/Constants.java
```diff
@@ -13,6 +13,10 @@
   // File Permission
   public static final FsPermission perm770 = new FsPermission((short) 0770);
 
+  // History Server related constants
+  public static final String JOBS_SUFFIX = "jobs";
+  public static final String CONFIG_SUFFIX = "config";
```
erwa: I would just call these constants `JOBS` and `CONFIG` and remove the `_SUFFIX` part.

Feel free to address in a separate PR or as part of another PR.

PR #135, opened 2018-12-15
tony doesnt wait for chief worker to finish before stopping #134

### LINKED ISSUE #134: tony doesn't wait for chief worker to finish before stopping
We wait for all workers to finish before stopping the training job. However, `chief` is another type of job we should also wait for.

Related code snippet.
```
      if (numCompletedWorkerTasks.get() == numTotalWorkerTasks) {
        Utils.printWorkerTasksCompleted(numCompletedWorkerTasks, numTotalWorkerTasks);
        break;
      }
```

This should be configurable as well.

oliverhu:
Fix #134

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
@@ -86,7 +86,7 @@ public class TonyConfigurationKeys {
   public static final String TONY_APPLICATION_PREFIX = TONY_PREFIX + "application.";
 
   public static final String APPLICATION_NAME = TONY_APPLICATION_PREFIX + "name";
-  public static final String DEFAULT_APPLICATION_NAME = "TensorFlowApplication";
+  public static final String DEFAULT_APPLICATION_NAME = "TonyApplication";
 
   public static final String FRAMEWORK_NAME = TONY_APPLICATION_PREFIX + "framework";
   public static final String DEFAULT_FRAMEWORK_NAME = "tensorflow";
@@ -171,6 +171,10 @@ public class TonyConfigurationKeys {
   public static final String WORKER_PREFIX = TONY_PREFIX + "worker.";
   public static final String WORKER_TIMEOUT = WORKER_PREFIX + "timeout";
   public static final int DEFAULT_WORKER_TIMEOUT = 0;
+
+  // Job types that we don't wait to finish
+  public static final String IGNORED_JOBTYPES = TONY_APPLICATION_PREFIX + "ignored.jobtypes";
+  public static final String IGNORED_JOBTYPES_DEFAULT = "ps";
 
   // Training chief configuration
   public static final String CHIEF_PREFIX = TONY_PREFIX + "chief.";
--- a/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
@@ -179,8 +179,8 @@ public class TonyConfigurationKeys {
   public static final int DEFAULT_WORKER_TIMEOUT = 0;
 
   // Job types that we don't wait to finish
-  public static final String IGNORED_JOBTYPES = TONY_APPLICATION_PREFIX + "ignored.jobtypes";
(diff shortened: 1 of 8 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
```diff
@@ -178,6 +178,10 @@ public static String getContainerResourcesKey() {
   public static final String WORKER_TIMEOUT = WORKER_PREFIX + "timeout";
   public static final int DEFAULT_WORKER_TIMEOUT = 0;
 
+  // Job types that we don't wait to finish
+  public static final String IGNORED_JOBTYPES = TONY_APPLICATION_PREFIX + "ignored.jobtypes";
+  public static final String IGNORED_JOBTYPES_DEFAULT = "ps";
```
erwa: Can we rename this `tony.untracked.jobtypes`?

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
```diff
@@ -251,4 +251,26 @@ public void testTonyResourcesFlag() throws ParseException {
...
+   * is gonna fail.
+   */
+  @Test
+  public void testWaitingForChiefToFinish() throws ParseException {
+    conf.setBoolean(TonyConfigurationKeys.IS_SINGLE_NODE, false);
+    conf.setInt(TonyConfigurationKeys.TASK_MAX_MISSED_HEARTBEATS, 2);
+    client = new TonyClient(conf);
+    client.init(new String[]{
+        "--src_dir", "tony-core/src/test/resources/",
+        "--executes", "tony-core/src/test/resources/conditional_wait.py",
```
erwa: After https://github.com/linkedin/TonY/pull/131, we should be able to just specify `--executes conditional_wait.py`, no?

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
```diff
@@ -251,4 +251,26 @@ public void testTonyResourcesFlag() throws ParseException {
     int exitCode = client.start();
     Assert.assertEquals(exitCode, 0);
   }
+
+
+  /*
+   * In conditional_wait.py script, we wait for 5s for chief job type. If we only wait for worker to finish, this test
+   * is gonna fail.
```
erwa: I don't understand why this test would fail even if we only waited for non-chief workers to finish.

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
oliverhu: Yah, i branched off master for this fix. didn't pick up #131 . will rebase that.

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
oliverhu: This test tries to mimic the scenario happens in #134. 

Prior to this change, we don't wait for non-chief worker to finish and conclude the training. The logic in TonySession's `updateSessionStatus()` method used to make decision based on if non-ps worker has finished, if no -> setFinalStatus false.

erwa: Got it, that makes sense. Thanks for explaining.

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
```diff
@@ -251,4 +251,26 @@ public void testTonyResourcesFlag() throws ParseException {
...
+   */
+  @Test
+  public void testWaitingForChiefToFinish() throws ParseException {
+    conf.setBoolean(TonyConfigurationKeys.IS_SINGLE_NODE, false);
+    conf.setInt(TonyConfigurationKeys.TASK_MAX_MISSED_HEARTBEATS, 2);
```
erwa: Are either of these two confs necessary? IS_SINGLE_NODE=false is already the default, and the missed heartbeats setting doesn't seem relevant to this test.

oliverhu: Makes sense. Removed those 2

PR #137, opened 2018-12-17
Minor refactoring

pdtran3k6:
Addressing https://github.com/linkedin/TonY/pull/107#discussion_r242214374, https://github.com/linkedin/TonY/pull/107#discussion_r242215375, and https://github.com/linkedin/TonY/pull/107#discussion_r242215810.
Updated testParserUtils + testHdfsUtils with Constants.TONY_CORE_SRC
Updated app config
Modified Parser method to use Java 8 API

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/util/ParserUtils.java
+++ b/tony-core/src/main/java/com/linkedin/tony/util/ParserUtils.java
@@ -203,15 +203,7 @@ public class ParserUtils {
   }
 
   public static List<JobEvent> mapEventToJobEvent(List<Event> events) {
-    return events.stream().map(ParserUtils::initJobEvent).collect(Collectors.toList());
-  }
-
-  private static JobEvent initJobEvent(Event e) {
-    JobEvent wrapper = new JobEvent();
-    wrapper.setType(e.getType());
-    wrapper.setEvent(e.getEvent());
-    wrapper.setTimestamp(e.getTimestamp());
-    return wrapper;
+    return events.stream().map(JobEvent::convertEventToJobEvent).collect(Collectors.toList());
   }
 
   private ParserUtils() { }
--- a/tony-core/src/main/java/com/linkedin/tony/util/ParserUtils.java
+++ b/tony-core/src/main/java/com/linkedin/tony/util/ParserUtils.java
@@ -16,6 +16,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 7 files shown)


tony-core/src/main/java/com/linkedin/tony/util/ParserUtils.java
```diff
@@ -202,15 +203,15 @@ public static JobMetadata parseMetadata(FileSystem fs, Path jobFolderPath, Strin
...
-      wrapper.setType(e.getType());
-      wrapper.setEvent(e.getEvent());
-      wrapper.setTimestamp(e.getTimestamp());
-      jobEvents.add(wrapper);
-    }
-    return jobEvents;
+    return events.stream().map(ParserUtils::initJobEvent).collect(Collectors.toList());
+  }
+
+  private static JobEvent initJobEvent(Event e) {
```
oliverhu: this looks weird. Look like it should be a method in JobEvent class?

pdtran3k6: Ahh I see. That makes more sense actually. Will update soon.

pdtran3k6:
Regarding https://github.com/linkedin/TonY/pull/107#issuecomment-447902575, I'm assuming by final folder you meant `intermediate` folder and _not_ the `finished` folder. 
With that being said, I think what we can do is we can write a temp file with the naming convention something like what you said (`<appId>-<startTime>-INPROGRESS-<user>.jhist`), and then we can modify the logic to just move the completed file (`<appId>-<startTime>-<completeTime>-<user>-<status>.jhist`) to `finished` and leave the inprogress one in `intermediate`. Wdyt? @erwa @oliverhu

oliverhu:
not sure what @erwa meant, I'll leave to him to comment

erwa:
> Regarding [#107 (comment)](https://github.com/linkedin/TonY/pull/107#issuecomment-447902575), I'm assuming by final folder you meant `intermediate` folder and _not_ the `finished` folder.

Yeah, I mean the `intermediate/<appId>` folder.

> With that being said, I think what we can do is we can write a temp file with the naming convention something like what you said (`<appId>-<startTime>-INPROGRESS-<user>.jhist`), and then we can modify the logic to just move the completed file (`<appId>-<startTime>-<completeTime>-<user>-<status>.jhist`) to `finished` and leave the inprogress one in `intermediate`. Wdyt? @erwa @oliverhu

I think we can write the temp INPROGRESS file to `intermediate/<appId>` (we may want to add a `.tmp` suffix to emphasize this is a temporary file that will be renamed -- this is what MR does, I believe). If the THS sees an INPROGRESS file, it can still show it on the job listings page but shouldn't move it to the `finished` directory.

When the application finishes, the TonyAM can rename the INPROGRESS file to SUCCEEDED/FAILED and add the finish time to the file name. (This is a rename within the same `intermediate/<appId>` folder.)

When the THS sees a SUCCEEDED/FAILED jhist file, then it can move the entire `<appId>` directory to the `finished` directory.

Issue #98, opened 2018-11-20
TonY Portal should enforce retention on history files


erwa:
Currently, history files are retained forever. The retention period should be configurable and TonY Portal should take care of enforcing retention.

As part of retention, we can also clean up in-progress files that are older than the retention period. (These are probably jobs that crashed or encountered other abnormal conditions.)

Internal Jira: LIHADOOP-43855

PR #140, opened 2018-12-18
Write to inprogress file

pdtran3k6:


tony-core/src/main/java/com/linkedin/tony/TonyJobMetadata.java
```diff
@@ -5,31 +5,28 @@
...
 
 
 public class TonyJobMetadata {
-  private static final Log LOG = LogFactory.getLog(TonyJobMetadata.class);
   private String id;
   private String url;
-  private long started;
-  private long completed;
+  private Long started;
+  private Long completed;
```
erwa: Why did you change these from `long` to `Long`? It's preferable to keep everything as a primitive `long` to avoid the implicit boxing/unboxing.

You probably changed them to object `Long` to support `null`, but another way to handle this is to use `-1L` to indicate a job does not have a start or end time.

tony-core/src/main/java/com/linkedin/tony/TonyJobMetadata.java
```diff
@@ -5,31 +5,28 @@
...
   private String user;
 
   // for testing only
   TonyJobMetadata() {
     this.id = "";
     this.url = "";
-    this.started = 0;
-    this.completed = 0;
+    this.started = (long) 0;
+    this.completed = (long) 0;
```
erwa: Rather than casting an int to a long, you can create a long literal by using `0L`.

tony-core/src/test/java/com/linkedin/tony/TestEventHandler.java
```diff
@@ -53,12 +53,13 @@ public void setup() {
   }
 
   @Test
```
erwa: Generally, you put the test for a class in the same package as the class itself. So since EventHandler is in package `com.linkedin.tony.events`, TestEventHandler should be in the same package under the test folder. By doing this, you can make the `@VisibleForTesting` methods package-private instead of public.

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -24,18 +23,26 @@
   private static final Log LOG = LogFactory.getLog(EventHandler.class);
   private boolean isStopped = false;
   private BlockingQueue<Event> eventQueue;
-  private Path historyFile = null;
-  private Path tmpFile = new Path(Constants.TMP_AVRO);
+  private Path finalHistFile = null;
+  private Path tmpHistFile;
```
erwa: Instead of `tmpHistFile`, can we call this `inProgressHistFile`?

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -86,25 +93,33 @@ public void run() {
...
+    // If setupThread method fails to create tmpHistFile,
+    // return immediately since we don't have any file to begin with
+    if (tmpHistFile == null) {
+      return;
+    }
+
+    // At this point, finalHistFile should be set
     // If not, then discard all events
-    if (historyFile == null) {
+    if (finalHistFile == null) {
```
erwa: Even if finalHistFile is not set, I don't think we should just delete the written events. Just leave them as is and they can get cleaned up during retention (#98). In fact, the written events should help us debug why the finalHistFile wasn't set (seeing which events have been written can help us pinpoint when in the application execution it crashed).

tony-core/src/test/java/com/linkedin/tony/TestHistoryFileUtils.java
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright 2018 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.tony;
```
erwa: Put the test for HistoryFileUtils in the same package as HistoryFileUtils, i.e. `com.linkedin.tony.util`.

tony-core/src/main/java/com/linkedin/tony/TonyJobMetadata.java
```diff
@@ -38,21 +35,26 @@ private TonyJobMetadata(String id, String url, long started, long completed, Str
     this.user = user;
   }
 
-  public static TonyJobMetadata newInstance(Configuration yarnConf, String appId, long started, long completed,
-      boolean status, String user) {
-    String jobStatus = status ? "SUCCEEDED" : "FAILED";
+  // Since newInstance could be called when the job hasn't finished,
+  // `completed` and `status` could be null
+  public static TonyJobMetadata newInstance(Configuration yarnConf, String appId, Long started, Long completed,
```
erwa: Another pattern to consider is using a Builder to construct a TonyJobMetadata. Example:

```
// After defining a static class Builder class inside TonyJobMetadata
new TonyJobMetadata.Builder().setConf(new Configuration()).setAppId(...).setStartTime(...).setStatus().setUser().build();
```

If something isn't set (e.g.: endTime), then it can default to some value like `-1L`. One advantage of this approach is it's less error-prone to accidentally swapping the arguments (e.g.: swapping start and end time or appId and user). You also avoid code like `TonyJobMetadata.newInstance(yarnConf, appIdString, started, null, null, user)`, from which you can't tell from a glance what the `null`s mean.

pdtran3k6:
@erwa: this is a quick fix, we can discuss more about the reason why the tests failed and possible solutions offline.

tony-core/src/main/java/com/linkedin/tony/TonyJobMetadata.java
pdtran3k6: 👍

tony-core/src/main/java/com/linkedin/tony/TonyJobMetadata.java
pdtran3k6: Yeah we can do that too.

tony-core/src/test/java/com/linkedin/tony/TestHistoryFileUtils.java
pdtran3k6: By this logic, I'm assuming all the other utils test (TestUtils, TestHdfsUtils, TestParserUtils, etc.) should also be moved to `test.com.linkedin.tony.util`?

erwa: Package name should not be prefixed `test.` (I think you're probably referring to `src/test/java/com...`), but yes, that's the general pattern.

PR #142, opened 2018-12-20
RM Link in THS

pdtran3k6:


tony-core/src/main/java/com/linkedin/tony/models/JobMetadata.java
```diff
@@ -37,7 +41,8 @@ public static JobMetadata newInstance(String histFileName) {
     long completed = Long.parseLong(metadata[2]);
     String user = metadata[3];
     String status = metadata[4];
-    return new JobMetadata(id, jobLink, configLink, started, completed, status, user);
+    String rmLink = Utils.buildRMUrl(new YarnConfiguration(), id);
```
erwa: Instead of creating a `new YarnConfiguration` each time `JobMetadata.newInstance()` is called, can we just pass in a `Configuration` object to this method instead?

tony-core/src/test/java/com/linkedin/tony/util/TestParserUtils.java
```diff
@@ -56,13 +57,15 @@ public void testIsValidHistFileName_false() {
   public void testParseMetadata_success() {
     Path jobFolder = new Path(Constants.TONY_CORE_SRC + "test/resources/typicalHistFolder/job1");
     String jobRegex = "application\\d+";
+    String RMLink = Utils.buildRMUrl(new YarnConfiguration(), "application123");
     JobMetadata expected = new JobMetadata("application123", "/" + Constants.JOBS_SUFFIX + "/application123",
-        "/" + Constants.CONFIG_SUFFIX + "/application123", 1, 1, "SUCCEEDED", "user1");
+        "/" + Constants.CONFIG_SUFFIX + "/application123", RMLink,1, 1, "SUCCEEDED", "user1");
```
erwa: Add space after `RMLink,`

tony-history-server/app/views/tableMetadata.scala.html
```diff
@@ -8,6 +8,7 @@
     <th>Completed</th>
     <th>User</th>
     <th>Status</th>
+    <th>RM</th>
```
erwa: How about `ResourceManager Link`?

tony-core/src/test/java/com/linkedin/tony/util/TestParserUtils.java
pdtran3k6: 👍

tony-core/src/main/java/com/linkedin/tony/models/JobMetadata.java
pdtran3k6: Sure. Also, I'm assuming that this is to limit the number of `YarnConfiguration` objects created?

tony-history-server/app/hadoop/Configuration.java
```diff
@@ -20,20 +20,25 @@
   private static final String CORE_SITE_CONF = YarnConfiguration.CORE_SITE_CONFIGURATION_FILE;
   private static final String HDFS_SITE_CONF = "hdfs-site.xml";
 
-  private static HdfsConfiguration conf = new HdfsConfiguration();
+  private static HdfsConfiguration hdfsConf = new HdfsConfiguration();
+  private static YarnConfiguration yarnConf = new YarnConfiguration();
```
erwa: Can we just use a single `Configuration` that we use for both HDFS and YARN stuff? If we switch to `Configuration`, I think we might need to add `yarn-site.xml` to the resources loaded (YarnConfiguration probably looks for and loads `yarn-site.xml` automatically, but I don't think `Configuration` does).

tony-core/src/main/java/com/linkedin/tony/models/JobMetadata.java
erwa: Yes, and for consistency. In case we want to add something to the `YarnConfiguration` in all of THS, we should be able to do it in one place rather than have to do it in every place we create a `new YarnConfiguration` instance.

PR #148, opened 2019-01-02
LIHADOOP-43241: TensorBoard not loading due to TB port not being released

### LINKED ISSUE #147: TensorBoard is not loading when using Estimator API with chief worker
In TaskExecutor.java, we only close the TB port we allocate if the task is a worker:

```
    if (executor.taskIndex == 0 && executor.jobName.equals("worker")) {
      executor.registerTensorBoardUrl();
      executor.tbSocket.close();
    }
```

However, with the new Estimator API, we should also close the port if it a "chief" task (chief worker), as we would then run TensorBoard in the Python process on the chief task.

erwa:
Fixed #147.


hungj:
LGTM

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
```diff
@@ -362,10 +362,8 @@ private TonyTask getTask(String jobName, String taskIndex) {
     return null;
   }
 
-  private boolean isChief(String jobName, String jobIndex) {
-    String chiefName = tonyConf.get(TonyConfigurationKeys.CHIEF_NAME, TonyConfigurationKeys.DEFAULT_CHIEF_NAME);
-    String chiefIndex = tonyConf.get(TonyConfigurationKeys.CHIEF_INDEX, TonyConfigurationKeys.DEFAULT_CHIEF_INDEX);
-    return jobName.equals(chiefName) && jobIndex.equals(chiefIndex);
+  private boolean isChief(String jobName) {
```
oliverhu: this part is not right. By default, worker 0 is chief worker, and we have logic that if chief is dead, we stop training (fix the hanging training scenario when chief worker is head and other workers are waiting)

erwa: I see what you're getting at. I think we need to modify the `TestTonyE2E.testWaitingForChiefToFinish` test, since it still passed after my changes :).

erwa: I added back the worker 0 check. However, I don't think the `TestTonyE2E.testWaitingForChiefToFinish` test was actually testing short-circuiting the training if the chief worker fails, so I removed it.

Issue #151, opened 2019-01-07
TonY History Server should support different filesystem for history files


erwa:
For federated HDFS instances, one may wish to write the history files to a separate filesystem than other logs. TonY History Server should support this.

PR #152, opened 2019-01-07
TonY History Server should support different filesystem for history files #151

erwa:
* Added `historyFs` to TonyAM that is used to write history files (can be different from `fs` used to load resources)
* Fixed `@Inject` setup in TonyHistoryServer. Previously, many injected instances were not being passed into constructors.
* Removed `Security` class and moved Kerberos setup into `Requirements()` constructor.
* Minor changes to `startTHS.sh`
* Added default `TONY_CONF_DIR` setting of `/export/apps/tony/conf`


tony-core/src/main/java/com/linkedin/tony/Constants.java
```diff
@@ -67,6 +67,7 @@
   public static final String HADOOP_CONF_DIR = ApplicationConstants.Environment.HADOOP_CONF_DIR.key();
   public static final String TONY_SITE_CONF = "tony-site.xml";
   public static final String TONY_CONF_DIR = "TONY_CONF_DIR";
+  public static final String DEFAULT_TONY_CONF_DIR = "/export/apps/tony/conf";
```
oliverhu: it is weird to put this as the default value

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -98,6 +98,7 @@
   private String appIdString;
   private String amHostPort;
   private FileSystem fs;
+  private FileSystem historyFs;  // FileSystem used to write history-related files like config and events
   private String tonyHistoryFolder;
   private Path jobDir = null;
   private String user = null;
```
oliverhu: This naming is confusing, what user is this and what for?

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -98,6 +98,7 @@
   private String appIdString;
   private String amHostPort;
   private FileSystem fs;
+  private FileSystem historyFs;  // FileSystem used to write history-related files like config and events
```
oliverhu: is this used for writing to another namenode?

tony-history-server/startTHS.sh
```diff
@@ -179,9 +179,10 @@ save () {
 }
 
 # Set up TONY_CONF_DIR if not exist
+DEFAULT_TONY_CONF_DIR=/export/apps/tony/conf
```
oliverhu: let's not put internal conf as default

tony-history-server/app/hadoop/Requirements.java
```diff
@@ -41,25 +50,57 @@ private void createDirIfNotExists(FileSystem myFs, Path dir, FsPermission perm)
...
-      return;
+  private void initFs(HdfsConfiguration conf) {
+    try {
+      histFs = histFolder.getFileSystem(conf);
+    } catch (IOException e) {
+      LOG.error("Encountered exception while getting filesystem for " + histFolder, e);
     }
+  }
+
+  private void setUpKeytab(HdfsConfiguration hdfsConf) {
```
oliverhu: minor: setup, to be consistent with below

tony-core/src/main/java/com/linkedin/tony/Constants.java
erwa: Currently, the `TONY_CONF_DIR` environment variable is not set by default. It might be a while until Grid SRE sets this on all client machines by default. Until then, I want users to automatically pick up `/export/apps/tony/conf/tony-site.xml` (which I've installed manually for now), so that history files get written out to the correct location.

Note that even if `/export/apps/tony/conf` does not exist on the client machine, there will not be any runtime exceptions.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
erwa: In our case, we want to write history files to NN02; that's what this `historyFs` is for. I'll expand the comment.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
erwa: Yes. I'll update the comment to make it more clear.

tony-history-server/startTHS.sh
erwa: Ok, I'll change the default to `$APP_HOME/conf`.

tony-history-server/app/hadoop/Requirements.java
erwa: Fixed

erwa:
Addressed Keqiu's comments.

erwa:
Added `TODO` comment to remove `DEFAULT_TONY_CONF_DIR` once `TONY_CONF_DIR` environment variable gets set globally on all our client machines.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
oliverhu: Sorry, I meant the `user` variable..

Issue #157, opened 2019-01-08
TonYApplicationMaster has race condition where thread is interrupted while flush is happening


erwa:
There's a race condition where the following might happen:

```
Main thread                      EventHandler thread
===========                      ===================

emitEvent(APPLICATION_FINISHED)
                                 take()
isStopped = true
                                 dataFileWriter.close()
interrupt()
``` 

`.close()` internally calls `.flush()`, and if the `interrupt()` happens during the `.flush()`, you sometimes get

```
2019-01-07 20:54:33 ERROR EventHandler:117 - Failed to close writer
java.io.InterruptedIOException: Interrupted while waiting for data to be acknowledged by pipeline
	at org.apache.hadoop.hdfs.DataStreamer.waitForAckedSeqno(DataStreamer.java:886)
	at org.apache.hadoop.hdfs.DFSOutputStream.flushInternal(DFSOutputStream.java:749)
	at org.apache.hadoop.hdfs.DFSOutputStream.closeImpl(DFSOutputStream.java:846)
	at org.apache.hadoop.hdfs.DFSOutputStream.close(DFSOutputStream.java:818)
	at org.apache.hadoop.fs.FSDataOutputStream$PositionCache.close(FSDataOutputStream.java:72)
	at org.apache.hadoop.fs.FSDataOutputStream.close(FSDataOutputStream.java:106)
	at java.io.FilterOutputStream.close(FilterOutputStream.java:159)
	at java.io.FilterOutputStream.close(FilterOutputStream.java:159)
	at org.apache.avro.file.DataFileWriter.close(DataFileWriter.java:446)
	at com.linkedin.tony.events.EventHandler.run(EventHandler.java:112)
```

and end up with a size-0 events file.

PR #158, opened 2019-01-08
Fix race condition where thread is interrupted while flush is in progress

erwa:
This PR fixes the race condition (see #157) by moving the `dataFileWriter.close()` call into the `stop()` method, after the event handler thread has terminated.


erwa:
Can't really think of a good, reliable way to test this.

PR #175, opened 2019-02-07
Publish fat jar to maven

hungj:


erwa:
Fat jar should already be in `tony-cli/build/libs` after building. We should publish that.

hungj:
Indeed. Fixed

erwa:
Actually, on further thought, we should publish **all** of the TonY jars, fat and thin. We should also publish the tony-history-server.zip Users may want to use the tony-azkaban jobtype plugin jar or run the tony-history-server, for example. Currently, we're not publishing these artifacts, but we should.

PR #176, opened 2019-02-07
Bump version to 0.2.0

hungj:


build.gradle
```diff
@@ -72,7 +72,7 @@ ext.deps = [
 
 allprojects {
   group = "com.linkedin.tony"
-  project.version = "0.1.5"
+  project.version = "0.1.6"
```
oliverhu: 0.2.0 to match internal versions. There are also a bunch of breaking changes between current head & previous releases.

hungj: Fixed

erwa: One such breaking change is https://github.com/linkedin/TonY/commit/073cfabba7ebc12fab37243ecf82cfb66ea72813 which changed the syntax from `--src_dir foo/bar --executes foo/bar/qux.py` to `--src_dir foo/bar --executes qux.py`.

Issue #180, opened 2019-02-08
Expand TonY --src_dir to take multiple directories


oliverhu:
As part of Az on TonY initiative, it would be easier if we could support passing in multiple source directories to be localized to all machines.

oliverhu:
Plan is to extend TonY's ClusterSubmitter to have another parameter of taking multiple directories, zip them, upload to HDFS and then append resources to tonyconf. Thought of adding this to TonyClient to expand, but I think `--resources` is generic enough.

erwa:
What about adding a `--src_dirs` argument that accepts a comma-separated list of directories?

oliverhu:
@erwa end goal is to deprecate --src_dirs in TonyClient and move this to submitter. --resources would be the only api for Tony client to accept resources.

Another thing I found useful to do is to add attributes to `--resources`:
--resources hdfs://abc/a.zip::archive,hdfs://def/g/::file (::file is optional), with this, we could leverage YARN to localize resources with cache.

PR #187, opened 2019-02-15
 Expand TonY --src_dir to take multiple directories #180

oliverhu:
Update:
- Added tests.
- New format: hdfs://user/hu/test.zip#archive It supports both local #archive and remote #archive, for example: /home/khu/abc#archive & hdfs://user/hu/test.zip#archive  both work.
- Updated wiki.

---
Initial PR to support local files and add :_:archives_ notation in _tony.XXX.resources_.
```
<property>
    <description>Name of your YARN application.</description>
    <name>tony.containers.resources</name>
    <value>/user/hu/helloworld.zip,hdfs://user/hu/remote.zip</value>
  </property>
```
Impl behind it:
We parse the tony.xxx.resources in tony conf, upload local files to remote fs and replace the local file entry with remote fs entries.

New annotation in tony.XX.resources:
```
hdfs://user/hu/test.zip::archive 
```
if ::archive is specified, the file will be unzipped when localized to the containers.


tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
```diff
@@ -8,19 +8,20 @@
 import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
-import java.util.Arrays;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
```
hungj: can we avoid using * import?

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -489,9 +500,27 @@ public static void createDir(FileSystem fs, Path dir, FsPermission permission) {
...
+    String dstAddress = dst.toString();
+    if (resourceType == LocalResourceType.ARCHIVE) {
+      dstAddress += "::archive";
+    }
+    appendConfResources(TonyConfigurationKeys.getContainerResourcesKey(), dstAddress, tonyConf);
+  }
+
+  public static void appendConfResources(String key, String resource, Configuration tonyConf) {
+    String currentResources = tonyConf.get(key, "");
+    tonyConf.set(key, currentResources + "," + resource);
```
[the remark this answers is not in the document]
oliverhu: Right, this is not correct. I should use
```
String[] resources = tonyConf.getStrings("')
append new resource to resources
tonyConf.setResources(key, resources)
```

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -431,6 +431,17 @@ public static void addResources(String[] resources, Map<String, LocalResource> r
   public static void addResource(String path, Map<String, LocalResource> resourcesMap, FileSystem fs) {
     try {
       if (path != null) {
+        // Check the format of the path, if the path is of path::archive, we set resource type as ARCHIVE
+        String[] pathType = path.split("::");
+        if (pathType.length == 2 && pathType[1].equals("archive")) {
+          assert Utils.isArchive(pathType[0]); // Assert it is an arvhive file.
```
erwa: `Utils.isArchive()` currently only works for local archive files. Can we also support remote archive files, e.g.: zips that are already on HDFS?

A few other minor things:
* I think `Preconditions.assertTrue(...)` is more idiomatic than `assert ...`
* `arvhive` --> `archive`

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -431,6 +431,17 @@ public static void addResources(String[] resources, Map<String, LocalResource> r
   public static void addResource(String path, Map<String, LocalResource> resourcesMap, FileSystem fs) {
     try {
       if (path != null) {
+        // Check the format of the path, if the path is of path::archive, we set resource type as ARCHIVE
+        String[] pathType = path.split("::");
```
erwa: This is a preference thing, but how about using `#archive` instead of `::archive`? MR uses `#alias` to support automatically renaming resource files, so I thought using `#archive` seemed more in line with the existing `#` delimiter pattern.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -431,6 +431,17 @@ public static void addResources(String[] resources, Map<String, LocalResource> r
...
+        if (pathType.length == 2 && pathType[1].equals("archive")) {
```
erwa: How about .equalsIgnoreCase to be more flexible?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -386,6 +387,53 @@ public void initTonyConf(Configuration tonyConf, CommandLine cliParser) throws I
...
   }
 
+  /**
+   * This helper method parses and updates tonyConf's jobtype.resources fields
+   *  - If the file is remote, keep as is.
+   *  - If the file is local and is a file, upload to remote fs and replace this entry with address of uploaded file.
+   *  - If the file is local and is a directory, zip the directory, upload to remote fs and replace entry with the
+   *  address of the uploaded file.
+   **/
+  private void processTonyConfResources(Configuration tonyConf, FileSystem fs) throws IOException {
```
erwa: Can we add some unit tests for this method?

oliverhu: Yah, wanted to throw out the idea and make sure the API is fine before adding those tests.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
oliverhu: makes sense

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
oliverhu: makes sense

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -489,9 +500,33 @@ public static void createDir(FileSystem fs, Path dir, FsPermission permission) {
...
+      dstAddress += Constants.ARCHIVE_SUFFIX;
+    }
+    appendConfResources(resourceKey, dstAddress, tonyConf);
+  }
+
+  public static void appendConfResources(String key, String resource, Configuration tonyConf) {
+    String[] resources = tonyConf.getStrings(key);
+    List<String> updatedResources = new ArrayList<>();
+    if (resources != null) {
+      updatedResources = new ArrayList<>(Arrays.asList(resources));
```
erwa: I think the `new ArrayList<>()` wrapping is unneeded.

oliverhu: It is needed because Array.asList returns an immutable list.

oliverhu:
Weird.. tests passed locally, need to double check.

oliverhu:
Figured out it is because we had zip files excluded. Updated .gitignore to not exclude files in resources/ folder

Issue #195, opened 2019-02-26
NotebookSubmitter hangs


erwa:
Currently, the NotebookSubmitter hangs because it waits for the notebook job to finish before starting the client proxy server. There is an extra `client.start()` line in NotebookSubmitter:

```
    client.init(updatedArgs);
    client.start();  // this line should be removed
    Thread clientThread = new Thread(client::start);
    clientThread.start();
```

PR #196, opened 2019-02-26
Fix NotebookSubmitter hangs #195

erwa:
Tested in a cluster.


erwa:
Also updated code to automatically kill the notebook application when the client process is stopped.

Issue #198, opened 2019-03-07
TonY race condition between worker termination notification and heartbeat liveness checker unregistration


erwa:
Currently, we do not unregister a task from the heartbeat monitor in the `TonyApplicationMaster` until we receive the container completion notification for the task. If the container completion notification is delayed more than 25 seconds (our default liveness threshold), then the heartbeat monitor might mark the task as failed, causing the entire job to fail.

One easy way to solve this race condition is to unregister the task from the heartbeat monitor before it terminates. This can be done in the registerExecutionResult() method the task calls before terminating.

PR #199, opened 2019-03-07
Fix TonY race condition between worker termination notification and heartbeat liveness checker unregistration #198

erwa:
Moved unregistration of task from heartbeat monitor into `registerExecutionResult()` to avoid the race condition.

Internal Jira: LIHADOOP-44694

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
@@ -996,14 +996,8 @@ public class TonyApplicationMaster {
     @Override
     public void onContainersCompleted(List<ContainerStatus> completedContainers) {
       LOG.info("Completed containers: " + completedContainers.size());
-      if (System.getenv(Constants.TEST_TASK_COMPLETION_NOTIFICATION_DELAYED) != null) {
-        LOG.info("Sleeping for 1 second to simulate task completion notification delay");
-        try {
-          Thread.sleep(1000);
-        } catch (InterruptedException e) {
-          LOG.error(e);
-        }
-      }
+      sleepForTesting();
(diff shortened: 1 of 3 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -984,6 +996,14 @@ public void onUpdateContainerResourceError(ContainerId containerId, Throwable t)
     @Override
     public void onContainersCompleted(List<ContainerStatus> completedContainers) {
       LOG.info("Completed containers: " + completedContainers.size());
+      if (System.getenv(Constants.TEST_TASK_COMPLETION_NOTIFICATION_DELAYED) != null) {
```
oliverhu: add some comments here.. this is a testing block

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
```diff
@@ -20,6 +20,8 @@
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
+import static com.linkedin.tony.TonyConfigurationKeys.*;
```
oliverhu: no star import

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
erwa: Extracted into a separate method.

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
erwa: Expanded.

PR #203, opened 2019-03-12
Fix bug when src_dir contains directory in the current directory or multiple directories

burgerkingeater:
This PR fixes error where TonyClient is not working when src directory is in the same directory of currently working directory or multiple directories is in src dir introduced in https://github.com/linkedin/TonY/pull/187.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -9,6 +9,7 @@ import azkaban.utils.Props;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.io.Files;
 import com.linkedin.tony.rpc.TaskUrl;
 import com.linkedin.tony.rpc.impl.ApplicationRpcClient;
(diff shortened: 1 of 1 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -463,7 +464,8 @@ public void processTonyConfResources(Configuration tonyConf, FileSystem fs) thro
                       fs, LocalResourceType.FILE, jobName);
             }
           } else {
-            Utils.zipFolder(Paths.get(srcDir), Paths.get(file.getName()));
+            File tmpDir = Files.createTempDir();
+            Utils.zipFolder(Paths.get(resource), Paths.get(tmpDir.getAbsolutePath(), file.getName()));
```
erwa: Now that you're creating the zip in `tmpDir`, I think you also need to update the `Utils. uploadFileAndSetConfResources()` call below to upload from the new location. Otherwise, it seems like the zip you created is not being used.

Can you also add a unit test for this?

oliverhu: That sounds right..  @burgerkingeater

oliverhu: How did it work when you tested it?

PR #209, opened 2019-03-14
Fix #208: TonY should prepend hdfs:// to hdfs_classpath if missing

### LINKED ISSUE #208: TonY should prepend hdfs:// to hdfs_classpath if missing
After https://github.com/linkedin/TonY/commit/81173476771d30d89329eb65a8a2924a8b05fa88#diff-6d0c97fa9e3e7630001399a4e18a218b, `hdfs_classpath` now requires a `hdfs://` prefix or else TonyClient will fail with `/my/path doesn't exist in local filesystem.`

TonyClient should automatically prepend `hdfs://` if missing.

erwa:
Ref: LIHADOOP-44864

Tested in Azkaban and also wrote a unit test.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -264,10 +264,18 @@ public class TonyClient implements AutoCloseable {
         + Utils.buildRMUrl(yarnConf, report.getApplicationId().toString()));
   }
 
-  private void createYarnClient() {
+  private void initHdfsConf() {
     if (System.getenv(Constants.HADOOP_CONF_DIR) != null) {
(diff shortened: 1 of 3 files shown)


oliverhu:
Shall we switch to use ClusterSubmitter in Azkaban? It would be nicer to keep TonyClient slim and move these flags/logics to cli classes.

erwa:
Yeah, that sounds like a good idea in general to unify the gateway and Azkaban launch paths. We might have to do some tweaks to the ClusterSubmitter to encompass the functionality of the CacheLibraryJob that most Azkaban flows start with.

erwa:
Filed https://github.com/linkedin/TonY/issues/210 for this.

PR #211, opened 2019-03-15
[Major Refactor] Update TonY to be compatible of Submarine runtime

oliverhu:
Refactor TonY to be more flexible and compatible with Submarine runtime interface.

- [x] Refactor TaskUrl to TaskStatus to record status of each task

- [x] Create TaskStatusListener to provide a way for users to track the statuses of each task from TonyClient.

- [x] Add API in TonyClient to invoke listener whenever the status changes.

- [x] Refactor the printTaskUrl logic 

- [x] Rename TonyApplicationMaster to ApplicationMaster

- [x] Make Docker image flexible for different job types.

- [x] Move container_env to TonyConf

- [x] Move executes to TonyConf

- [x] Move shell_env to TonyConf

- [x] Validation

- [x] Add tests

- [x] Test in internal environment


tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
```diff
@@ -256,8 +253,10 @@ public void onTaskCompleted(String jobName, String jobIndex, int exitCode) {
           if (isChief(jobName, jobIndex)) {
             trainingFinished = true;
           }
+          task.getTaskInfo().setState(TaskStatus.FAILED);
           setFinalStatus(FinalApplicationStatus.FAILED, "Exit status: " + exitCode);
         }
+        task.getTaskInfo().setState(TaskStatus.SUCCEEDED);
```
hungj: This doesn't seem right? Should this be in an else {} ?

tony-core/src/main/java/com/linkedin/tony/rpc/impl/TaskStatus.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+package com.linkedin.tony.rpc.impl;
+
+public enum TaskStatus {
+    NEW,
+    READY,
+    RUNNING,
+    FAILED,
+    SUCCEEDED;
+    TaskStatus() { }
```
erwa: Is this constructor necessary? I think it's not required.

tony-core/src/main/java/com/linkedin/tony/rpc/impl/pb/GetTaskInfosResponsePBImpl.java
```diff
@@ -0,0 +1,72 @@
+/**
+ * Copyright 2018 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
```
erwa: 2019

build.gradle
```diff
@@ -72,7 +72,7 @@ ext.deps = [
 
 allprojects {
   group = "com.linkedin.tony"
-  project.version = "0.2.0"
+  project.version = "0.3.0-SNAPSHOT"
```
erwa: Is the SNAPSHOT necessary?

tony-core/src/main/java/com/linkedin/tony/rpc/impl/TaskStatus.java
oliverhu: This is just make it not public, i think it is an IDE warning

build.gradle
oliverhu: This is temporary, will remove it, personal preference, don't want to mess up with maven central versions.

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
oliverhu: right..

tony-core/src/main/java/com/linkedin/tony/rpc/impl/TaskStatus.java
oliverhu: hmm. it seems i don't have to put it here. removed

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -718,6 +693,12 @@ private int doPreprocessingJob() throws Exception {
     extraEnv.put("HOME", System.getProperty("user.dir"));
     extraEnv.put(Constants.PY4JGATEWAY, String.valueOf(gatewayServerPort));
     String taskCommand = baseTaskCommand;
+    if (tonyConf.get(TonyConfigurationKeys.getExecuteCommandKey(Constants.AM_NAME),
+            tonyConf.get(TonyConfigurationKeys.getContainerExecuteCommandKey())) != null) {
+        taskCommand = tonyConf.get(TonyConfigurationKeys.getExecuteCommandKey(Constants.AM_NAME),
+                tonyConf.get(TonyConfigurationKeys.getContainerExecuteCommandKey()));
```
erwa: I think this `if` block would be more readable if you used a local variable, e.g.:
```
String amCommand = tonyConf.get(..., tonyConf.get(...));
if (amCommand != null) {
  taskCommand = amCommand;
}
```

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
```diff
@@ -204,7 +195,13 @@ public synchronized TonyTask getAndInitMatchingTask(long allocationRequestId) {
       TonyTask[] tasks = entry.getValue();
       for (int i = 0; i < tasks.length; i++) {
         if (tasks[i] == null) {
-          tasks[i] = new TonyTask(jobName, String.valueOf(i), sessionId);
+          String taskCommand = taskCmd;
+          if (tonyConf.get(TonyConfigurationKeys.getExecuteCommandKey(jobName),
+                  tonyConf.get(TonyConfigurationKeys.getContainerExecuteCommandKey())) != null) {
+            taskCommand = tonyConf.get(TonyConfigurationKeys.getExecuteCommandKey(jobName),
```
erwa: Same comment as above -- please use a local variable for readability and to avoid repeating yourself.

tony-core/src/test/java/com/linkedin/tony/TestTaskExecutor.java
```diff
@@ -4,45 +4,47 @@
...
     TaskExecutor taskExecutor = new TaskExecutor();
     Configuration tonyConf = new Configuration(false);
     tonyConf.setInt(TonyConfigurationKeys.TASK_HEARTBEAT_INTERVAL_MS, 2000);
     File confFile = new File(System.getProperty("user.dir"), Constants.TONY_FINAL_XML);
     try (OutputStream os = new FileOutputStream(confFile)) {
       tonyConf.writeXml(os);
     }
+    // Should throw exception since we didn't set up Task Command.
     taskExecutor.initConfigs(args);
     assertEquals(2000, taskExecutor.tonyConf.getInt(TonyConfigurationKeys.TASK_HEARTBEAT_INTERVAL_MS,
```
erwa: If `taskExecutor.initConfigs(args)` should throw an exception, please remove lines 43-47 as they won't be executed then.

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
```diff
@@ -35,14 +51,16 @@
  *
  * The YARN logs for the test should be in {@code <TonY>/target/MiniTonY}.
  */
-public class TestTonyE2E {
+public class TestTonyE2E implements CallbackHandler, TaskUpdateListener {
```
erwa: It doesn't seem clean to me for `TestTonyE2E` to directly implement these interfaces. I suggest creating a nested class that implements the callback interfaces.

tony-core/src/main/java/com/linkedin/tony/client/TaskUpdateListener.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+package com.linkedin.tony.client;
+
+import com.linkedin.tony.rpc.TaskInfo;
+
+import java.util.Set;
+
+public interface TaskUpdateListener {
+    // Called when TonyClient gets a set of taskUrls from TonyAM.
+    public void onTaskInfosReceived(Set<TaskInfo> taskInfoSet);
```
erwa: Can we combine `TaskUpdateListener` with `CallbackHandler` and just have one callback interface instead of two?

erwa:
I saw 1 failed test in Travis CI:

```
Gradle suite > Gradle test > com.linkedin.tony.TestTonyE2E.testTonyClientCallbackHandler FAILED
    java.lang.AssertionError: expected:<SUCCEEDED> but was:<RUNNING>
        at org.testng.Assert.fail(Assert.java:89)
        at org.testng.Assert.failNotEquals(Assert.java:489)
        at org.testng.Assert.assertEquals(Assert.java:118)
        at org.testng.Assert.assertEquals(Assert.java:160)
        at com.linkedin.tony.TestTonyE2E.testTonyClientCallbackHandler(TestTonyE2E.java:352)
```

Could it be a race condition?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
oliverhu: removed this logic, taskCommand is always in conf after previous changes

tony-core/src/test/java/com/linkedin/tony/TestTaskExecutor.java
oliverhu: moved logic below that to above that statement

tony-core/src/main/java/com/linkedin/tony/client/TaskUpdateListener.java
oliverhu: They are different, CallbackHandler is something set up in the constructor to receive all one time notifications. for TaskUpdateListener, it is more like checking a stream of updates.

tony-core/src/test/java/com/linkedin/tony/TestTaskExecutor.java
oliverhu: nvm.. removed

tony-core/src/main/java/com/linkedin/tony/client/TaskUpdateListener.java
erwa: I think we only call this callback once, when all the tasks have registered, so it seems a one-time notification as well.

Also, it seems that in your PR, no code is calling `onTaskInfosReceived`?

erwa: Nevermind, I couldn't find it via Ctrl-F because GitHub had automatically collapse the `TonyClient` changes because the diff was considered "large".

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -756,18 +811,31 @@ private boolean monitorApplication()
...
+                .filter(taskInfo -> !taskInfos.contains(taskInfo))
+                .collect(Collectors.toSet());
+        // If task status is changed, invoke callback for all listeners.
+        if (!taskInfoDiff.isEmpty()) {
+          LOG.info(receivedInfos);
+          for (TaskUpdateListener listener : listeners) {
+            listener.onTaskInfosReceived(receivedInfos);
+          }
+        }
+        taskInfos = receivedInfos;
```
erwa: Why do you need both `taskInfos` and `receivedInfos`?

oliverhu: Received info is what you received from last call to AM, taskInfos is the previous call to AM. Here is a comparison between these two and we only invoke the listener if anything has changed

tony-core/src/main/java/com/linkedin/tony/client/TaskUpdateListener.java
oliverhu: no.. every time the taskInfo is updated, we send notification

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
erwa: Ah, I see. I think this should be moved up one line to inside the `if (!taskInfoDiff.isEmpty())` block? If there is no diff, we don't need to update `taskInfos` I think. (I guess from a correctness standpoint it doesn't matter either way, but I just think logically it makes more sense to only update it if something actually changed.)

tony-core/src/main/java/com/linkedin/tony/client/TaskUpdateListener.java
erwa: True, the code calls `onTaskInfosReceived` every time there's a diff, but in practice, a change happens only one time:

```
// in ApplicationMaster.getTaskInfos()
      if (!singleNode && session != null && session.allTasksScheduled()) {
        return session.getTonyTasks().values().stream()
            .flatMap(tasks -> Arrays.stream(tasks).map(TonyTask::getTaskInfo))
            .collect(Collectors.toSet());
      }

      return Collections.emptySet();
```

We keep returning an empty set until ALL tasks have been scheduled, at which point we return a set containing all the tasks. There are no changes after that. So in effect, `onTaskInfosReceived` is a one-time callback.

oliverhu: not really. the state of the task changes, if the task is changed from NEW -> RUNNING -> SUCCEEDED / FAILED, the will be triggered

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
oliverhu: Makes sense, moved

tony-core/src/main/java/com/linkedin/tony/client/TaskUpdateListener.java
erwa: Are you sure? Can you add a print statement and see how many times onTaskInfosReceived() is called in a normal TonY job? Also, if it's intended that this method is called each time the task state changes, I think a better name would be `onTaskInfosChanged`.

oliverhu: yah.. only when changed. That is a good point, **onTaskInfosChanged** is more proper.

oliverhu:
Done testing internally..

PR #172, opened 2019-02-04
TonY working with Hadoop 2.7.2

anupam128:
Looking for some preliminary feedback.
We propose to create a separate branch for Hadoop 2.7 as these changes cannot co-exist in master.
Details of changes:
 * Fixed breaking changes in AM-RM/NM protocol
 * All new methods were unused
 * Ignore GPU with warning (should we fail instead?)
 * Block docker
 * Match based on container priority
 * Get createRetryPolicy to work
 * Deleted azkaban code as guice and guava version were incompatible with hadoop requirement (may be some gradle magic can fix this?)
 * All tests pass
 * Tested in Hadoop 2.7.3 cluster in distributed and pseudo-distributed mode.

---
This change is [Reviewable](https://reviewable.io/reviews/linkedin/tony/172)

PR #215, opened 2019-03-22
Fix #214: Build TonY against Hadoop 2.7 

### LINKED ISSUE #214: Build TonY against Hadoop 2.7
To support legacy clusters running Hadoop 2.7. Most of Hadoop 2.9+ APIs used can be called using reflection if available.

erwa:
Incorporated changes from https://github.com/linkedin/TonY/pull/172 plus added back GPU support and fixed some minor things.

Internal Jira: LIHADOOP-45062

@anupam128 I've incorporated your changes + some additional changes in this PR.


erwa:
I tested on both our Hadoop 2.7 and Hadoop 2.9 clusters and TonY works on both. Also, GPUs are still supported on the Hadoop 2.9 cluster.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -854,6 +889,89 @@ public int start() {
...
+   * correctly handles HA resource manager configurations.
+   *
+   * @param conf the {@link Configuration} file from which to read the
+   * principal
+   * @return the resource manager's principal string or null if the
+   * {@link YarnConfiguration#RM_PRINCIPAL} property is not set in the
+   * {@code conf} parameter
+   * @throws IOException thrown if there's an error replacing the host name
+   */
+  public static String getRmPrincipal(Configuration conf) throws IOException {
```
oliverhu: is this piece of code copied from Hadoop 2.9?

erwa: Yes, changes are from https://github.com/linkedin/TonY/pull/172.

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
```diff
@@ -137,10 +138,6 @@
   private AtomicInteger numRequestedContainers = new AtomicInteger();
   private Map<String, List<ContainerRequest>> jobTypeToContainerRequestsMap = new HashMap<>();
 
-  // allocationRequestIds are allocated to tasks sequentially. lastAllocationRequestId
-  // tracks the latest allocationRequestId allocated.
-  private long lastAllocationRequestId = 0;
```
oliverhu: If this is gone, how do we track the allocation request now?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -390,22 +390,57 @@ public boolean init(String[] args) throws ParseException, IOException {
...
+    }
+
+    return true;
+  }
+
+  /**
+   * Parses Docker related configs and sets the appropriate container environment variables if Docker is available.
+   * Uses reflection to support older versions of Hadoop.
+   */
+  private boolean parseDockerConfigs() {
```
oliverhu: This makes me think twice on merging this into master branch. There is gonna be more and more docker related featured been developed in Hadoop 3.x, this is gonna make future development miserable..

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
```diff
@@ -177,35 +174,24 @@ public boolean allTasksScheduled() {
...
-      jobTypeToAllocationIds.put(jobName, new HashSet<>());
-    }
-    jobTypeToAllocationIds.get(jobName).add(allocationRequestId);
-    LOG.info(String.format("Job %s with allocationRequestId %d", jobName, allocationRequestId));
-  }
-
   /**
    * Get a TensorFlow task that hasn't been scheduled.
-   * @param allocationRequestId the allocationRequestId of the allocated container
+   * In the absence of allocationRequestId, we are relying on the fact that each tensorflow job will
```
oliverhu: this is a bit too hacky

tony-core/src/main/java/com/linkedin/tony/TonyApplicationMaster.java
erwa: We match by priority -- see `TonySession.getAndInitMatchingTaskByPriority()`.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
erwa: I think we can make it maintainable by keeping Docker-related methods in one place -- perhaps creating a DockerUtils class.

Also, we already support Docker -- we just rely on YARN being configured to use the Docker container runtime and users specifying a Docker image via config.

What other Docker work is needed?

erwa: Another alternative if we want to keep building against a newer Hadoop version is to use the `hadoop27` branch to publish a separate 2.7-compatible internal TonY build that can run on our Hadoop 2.7 clusters. We could use version numbers to differentiate the 2, e.g.: 0.1.x for TonY Hadoop 2.7 releases and 0.2.x+ for TonY Hadoop 2.9 releases.

What do you think we should do?

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
erwa: Hadoop 2.7 doesn't have `allocationId` in `ContainerRequest` and `Container`. The only other way I can think of for matching containers to tasks is to do resource matching (match the memory and vcores returned to that requested for a task type). Do you prefer that?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
oliverhu: @erwa I think an Executor config or class to abstract out things would be useful. On a second thought, if we want to run TonY in 2.7 environment, it is much better to have single version to support both, otherwise, users would have to build two MPs one for adhoc cluster and one for prod cluster.

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
oliverhu: I think we used to use resources for matching, but that doesn't work for some cases (can't remember the details). @hungj thoughts?

erwa:
@hungj, @oliverhu, and I synced up offline. The consensus is to migrate master branch to Hadoop 2.7 (i.e., go forth with this PR). Could you guys take another look and approve this if it looks good?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
erwa: Yeah, we're in alignment that having a single version that supports both 2.7 and 2.9 makes more sense.

I think we can do an "Executor config" refactoring in a separate PR.

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
erwa: Then I guess we have no choice but to match by priority?

oliverhu:
LG overall, let me merge #211  first 🙏 🤣

erwa:
Hey @anupam128 ,

Thanks for the comments. Yeah, I agree relying on priority as a workaround for there not being allocation request id in Hadoop 2.7 is a bit hacky. To make it a bit less hacky for 2.9+, we could use reflection to check if allocationId is supported and use it if it is.

I think for now, we at LinkedIn prefer maintaining a single version that works on both 2.7 and 2.9 in order to support our legacy clusters until they fully migrate to 2.9+. In the future, once we've fully migrated to 2.9+, we can move the existing code to the hadoop27 branch and upgrade master to build against Hadoop 2.9+.

erwa:
Rebased on master.

erwa:
@oliverhu , can you re-review now that your patch is merged?

oliverhu:
There is another problem for this.. for Submarine integration, I don't think this would be compatible..

erwa:
@oliverhu , can you elaborate on why this code might not be compatible with Submarine? We should always be able to use reflection.

erwa:
We also need this patch in order to run on our legacy 2.7 production cluster.

oliverhu:
dependency mismatches, let me build Submarine and see how it goes

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -181,15 +176,23 @@ private ApplicationMaster() {
...
+    hbMonitor = new AbstractLivelinessMonitor<TonyTask>("Tony Task liveliness Monitor", new UTCClock()) {
       @Override
       protected void expire(TonyTask task) {
         onTaskDeemedDead(task);
       }
 
       @Override
       protected void serviceStart() throws Exception {
-        setMonitorInterval(hbInterval * 3);
+        // setMonitorInterval(int) changed to setMonitorInterval(long) in Hadoop 2.9,
```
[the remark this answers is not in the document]
erwa: This works for both 2.7 and 2.9 because it uses reflection to iterate over all methods in `AbstractLivelinessMonitor`. In 2.7, it will find `setMonitorInterval(int)`, and in 2.9, it will find `setMonitorInterval(long)`. Java will do automatic casting of int to long if the 2.9 `setMonitorInterval(long)` method is found.

PR #223, opened 2019-03-25
Update README.md

gogasca:
Add auto-scaling support to be able to launch multiple TensorFlow Jobs


erwa:
Hmm, I'm not sure how this would work with TonY and TensorFlow, since TensorFlow doesn't support modifying the graph what's it's finalized and the workers and parameter servers have started. Also, TonY only requests containers one time based on the initial configuration.

So I'm unclear how auto-scaling would work given the current behavior of TensorFlow and TonY.

oliverhu:
@gogasca  and I discussed in the tony slack channel, this is more about auto scaling the Hadoop cluster

tony-examples/tony-in-gcp/README.md
```diff
@@ -219,15 +219,93 @@ You can also track Job status from DataProc Jobs tab: Cloud Console -> Big Data
 
 ![img](https://i.imgur.com/3XgZI5Z.png)
 
+## Autoscaling
+
+
```
oliverhu: worth clarifying this is for autoscaling the Hadoop cluster, not each individual TF or TonY job

tony-examples/tony-in-gcp/README.md
```diff
@@ -219,15 +219,93 @@ You can also track Job status from DataProc Jobs tab: Cloud Console -> Big Data
...
 ![img](https://i.imgur.com/3XgZI5Z.png)
 
+## Autoscaling
+
+
+Estimating the "right" number of cluster workers (nodes) for a workload is difficult, and a single cluster size for an entire pipeline is often not ideal. User-initiated Cluster Scaling partially addresses this challenge, but requires monitoring cluster utilization and manual intervention.
+
+The Cloud Dataproc AutoscalingPolicies API provides a mechanism for automating cluster resource management and enables cluster autoscaling. An Autoscaling Policy is a reusable configuration that describes how clusters using the autoscaling policy should scale. It defines scaling boundaries, frequency, and aggressiveness to provide fine-grained control over cluster resources throughout cluster lifetime.
+
+Please find a sample below:
```
oliverhu: if I'm the user, it would be helpful to have a quick gif or video on this

erwa:
Ah, now I understand auto scaling is scaling the number of nodes in the cluster and not scaling an application. I got confused because the term "worker" is also one of the TensorFlow task types! It might help to add a sentence clarifying that this auto scales the number of nodes in your cluster, not the number of containers in an application instance.

tony-examples/tony-in-gcp/README.md
gogasca: Ack

tony-examples/tony-in-gcp/README.md
gogasca: Gif added

PR #225, opened 2019-03-26
Fix #224: Change application type from TENSORFLOW to TonY

### LINKED ISSUE #224: Change application type from TENSORFLOW to TonY
Currently, TonyClient sets the YARN application type to TENSORFLOW, but TonY can launch other types of programs, too. We should update the application type to TonY.

erwa:


tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -91,7 +91,7 @@
 public class TonyClient implements AutoCloseable {
   private static final Log LOG = LogFactory.getLog(TonyClient.class);
 
-  private static final String APP_TYPE = "TENSORFLOW";
+  private static final String APP_TYPE = "TonY";
```
oliverhu: TONY ?

erwa: Looks like MAPREDUCE and SPARK use ALL_CAPS, too. Will change to TONY.

PR #228, opened 2019-03-28
Add LocalizableResource class for parsing complicated path pattern

oliverhu:
Now support

hdfs://123.txt::abc.txt#archive
SOURCEFILE :: LOCALFILENAME :: TYPE


erwa:
Can you rebase on `master`? There's a conflict that can't be auto-resolved.

oliverhu:
mm, totally forgot about this PR, will rebase master

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
```diff
@@ -0,0 +1,90 @@
...
+        if (tuple.length == 2) {
+            localFileName = tuple[1];
+        }
+
+        if (sourceFileStatus.isDirectory()) {
+            isDirectory = true;
+        }
+    }
+
+    public LocalResource getLocalResource() {
```
erwa: `toLocalResource()` sounds more accurate to me (you're converting from one representation to another).

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
```diff
@@ -0,0 +1,90 @@
...
+    public LocalizableResource(String path) {
+        this.path = path;
+    }
+
+    public void parse(FileSystem fs) throws ParseException, IOException {
+        String filePath = path;
+        resourceType = LocalResourceType.FILE;
+        if (path.contains(Constants.ARCHIVE_SUFFIX)) {
+            resourceType = LocalResourceType.ARCHIVE;
+            filePath = path.replace(Constants.ARCHIVE_SUFFIX, "");
```
hungj: In addition to @erwa's path.endswith comment, 

path.substring(0, path.length() - Constants.ARCHIVE_SUFFIX.length())

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public String getLocalFileName() {
+        return localFileName;
+    }
+
+    public LocalizableResource(String path) {
+        this.path = path;
+    }
+
+    public void parse(FileSystem fs) throws ParseException, IOException {
```
hungj: Any reason this needs to be a separate method? Can we just make this private and call it in the constructor?

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
```diff
@@ -0,0 +1,90 @@
...
+        String filePath = path;
+        resourceType = LocalResourceType.FILE;
+        if (path.contains(Constants.ARCHIVE_SUFFIX)) {
```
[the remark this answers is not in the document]
oliverhu: updated to `endsWith`, i think case-insensitive is fine

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
oliverhu: ok.. this makes sense to use together with @erwa 's insensitive archive_suffix comment

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
oliverhu: that's a good point, updated

PR #238, opened 2019-04-15
Add sample scripts for Next presentation

gogasca:
- Add create_cluster script
- Run multiple jobs
- Installs GPU drivers (CUDA 9 and CUDA 10)
- Deploy TonY (Build .jar file and add samples)


erwa:
Great talk at Google Cloud NEXT '19, @gogasca! Thanks for adding these scripts here.

Issue #242, opened 2019-04-22
TonY should collect and write out metrics for all tasks


erwa:
One way to do this is to have the TaskExecutor monitor the task and periodically report metrics back to the AM. The AM can then write out the metrics for the task when the task finishes.

PR #243, opened 2019-04-22
TonY should collect and write out metrics for all tasks #242

erwa:
Internal Jira: LIHADOOP-45751

Overview of changes:

* Added a `TaskFinished` event which gets written by the AM when the task finishes
* Added a metrics RPC server to the AM and added a `TaskMonitor` to the `TaskExecutor` which will monitor the task and report metrics back to the AM (to the metrics RPC server) via a `MetricsRpc` protocol periodically
* Added monitoring of max memory usage of each task
* Added `MetricWritable` and `MetricsWritable` classes for serializing metrics over the wire using Hadoop's default WritableRpcEngine.


erwa:
The `TASK_FINISHED` event looks something like this:

```
 {
  "type" : "TASK_FINISHED",
  "event" : {
    "com.linkedin.tony.events.TaskFinished" : {
      "taskType" : "worker",
      "taskIndex" : 0,
      "status" : "SUCCEEDED",
      "metrics" : [ {
        "name" : "MEMORY_BYTES",
        "value" : 3.89177344E8
      } ]
    }
  },
  "timestamp" : 1555900943179
}
```

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
```diff
@@ -282,6 +313,48 @@ public void run() {
     }
   }
 
+  /**
+   * Monitors the task and reports metrics to the AM.
+   */
+  private class TaskMonitor implements Runnable {
```
oliverhu: move this out to a separate file ?

tony-core/src/main/resources/META-INF/services/org.apache.hadoop.security.SecurityInfo
```diff
@@ -1,3 +1,3 @@
 # Copyright 2018 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
 # See LICENSE in the project root for license information.
-com.linkedin.tony.TFClientSecurityInfo
+com.linkedin.tony.TonyClientSecurityInfo
```
oliverhu: Cool!

tony-core/src/main/java/com/linkedin/tony/TonyPolicyProvider.java
```diff
@@ -4,16 +4,20 @@
...
+import com.linkedin.tony.rpc.TensorFlowCluster;
 import org.apache.hadoop.security.authorize.PolicyProvider;
 import org.apache.hadoop.security.authorize.Service;
-import com.linkedin.tony.rpc.TensorFlowCluster;
 
 /**
  * PolicyProvider for Client to AM protocol.
  **/
-public class TFPolicyProvider extends PolicyProvider {
+public class TonyPolicyProvider extends PolicyProvider {
```
oliverhu: Thanks for the clean up! Could you write a doc on how to add a new RPC between AM & Executors?

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
```diff
@@ -182,7 +192,12 @@ protected void initConfigs(String[] args) throws Exception {
     Options opts = new Options();
     opts.addOption("am_address", true, "The address to the application master.");
     CommandLine cliParser = new GnuParser().parse(opts, args);
-    amAddress = cliParser.getOptionValue("am_address", "");
+
+    String amAddress = cliParser.getOptionValue("am_address", "");
```
oliverhu: can we also move this to be an environment variable to get rid of all argument passing ?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -402,25 +405,34 @@ private boolean prepare() {
...
-      }
-    } catch (Exception e) {
+      response = amRMClient.registerApplicationMaster(amHostname, amPort, null);
+      amHostPort = amHostname + ":" + amPort;
+    } catch (YarnException e) {
       LOG.error("Exception while preparing AM", e);
       return false;
     }
 
+    if (secureMode) {
```
oliverhu: Add a try catch block around this, currently if a cluster is not secure but I don't specify insecure mode, there will be a vague exception thrown out. We should explicitly tell the user the mistake

erwa: I let any `IOException` propagate and crash the AM. There should be a stack trace.

To really debug this when I was developing, I had to enable debug logging:
```
org.apache.log4j.Logger logger4j = org.apache.log4j.Logger.getRootLogger();
logger4j.setLevel(org.apache.log4j.Level.toLevel("DEBUG"));
```

or change the logging level in the tony-cli `log4j.properties`.

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
erwa: Good idea. Changed to environment variables `AM_HOST` and `AM_PORT` in my latest revision.

tony-core/src/main/java/com/linkedin/tony/TonyPolicyProvider.java
erwa: Sure, I'll add a Wiki page for this on the TonY GitHub Wiki.

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
erwa: Done

tony-core/src/main/java/com/linkedin/tony/TonyPolicyProvider.java
erwa: Created https://github.com/linkedin/TonY/wiki/Adding-an-RPC-client---server-pair-between-client-and-AM

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
oliverhu: The stack trace is like - failed to get RM token. We should add something average people could understand..

erwa: Hmm, I felt an error message didn't help much when developing; I had to enable DEBUG logging to really understand what was going on. That said, if I encounter issues with this code in the future and this part of the code is the culprit, I'll add some more logging then.

Issue #246, opened 2019-04-22
SRC_FILE::DEST_FILE#archive doesn't work in TonyClient


oliverhu:
This convention only works in AM, but doesn't seem to work in TonyClient 🤔

PR #249, opened 2019-04-23
Fix SRC_FILE::DEST_FILE#archive doesn't work in TonyClient #246

oliverhu:
SRC_FILE::DEST_FILE#archive doesn't work in TonyClient #246

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
+++ b/tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
@@ -35,6 +35,8 @@ public class LocalizableResource {
   private LocalResourceType resourceType;
   private FileStatus sourceFileStatus;
   private Path sourceFilePath;
+
+  // The file path inside the container.
   private String localFileName;
 
   private LocalizableResource() { }
@@ -62,6 +64,11 @@ public class LocalizableResource {
   public LocalizableResource(String path, FileSystem fs) throws ParseException, IOException  {
     this.path = path;
       this.parse(fs);
+  }
+
+  public void updateSourcePath(String sourcePath, FileSystem fs) throws IOException, ParseException {
+    path = sourcePath;
+    parse(fs);
   }
 
   private void parse(FileSystem fs) throws ParseException, IOException {
--- a/tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
+++ b/tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
@@ -42,6 +42,14 @@ public class LocalizableResource {
     return isDirectory;
   }
 
+  public boolean isLocalFile() {
+    return new Path(path).toUri().getScheme() == null;
+  }
+
+  public boolean isArchive() {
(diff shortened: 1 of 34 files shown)


tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
```diff
@@ -34,6 +35,8 @@
   private LocalResourceType resourceType;
```
erwa: What's the difference between `path` and `localFileName`? And how do they differ from `sourceFilePath`? Can you add some Javadoc explaining the differences?

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
```diff
@@ -68,8 +84,14 @@ private void parse(FileSystem fs) throws ParseException, IOException {
         throw new ParseException("Failed to parse file: " + path);
     }
 
+    FileSystem localFs = FileSystem.getLocal(new Configuration());
```
erwa: Please don't create `new Configuration()` here. Pass in a `Configuration` object to this method if needed.

Also, you can move this inside the `if` block on lines 90-92.

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
```diff
@@ -42,6 +45,14 @@ public boolean isDirectory() {
...
+    return new Path(path).toUri().getScheme() == null;
+  }
+
+  public boolean isArchive() {
+    return resourceType == LocalResourceType.ARCHIVE;
+  }
+
   public Path getSourceFilePath() {
     return sourceFilePath;
   }
```
erwa: (I wasn't able to comment directly on line 66, so I'm commenting here.) Indentation is off on line 66.

tony-core/src/test/resources/scripts/check_archive_file_localization.py
```diff
@@ -22,7 +22,9 @@
 if not os.path.isfile('./test.zip'):
     logging.error('test.zip doesn\'t exist')
     exit(-1)
```
erwa: Will this fail now that you've changed `tony-core/src/test/resources/test.zip` to `tony-core/src/test/resources/test.zip::test20.zip` in the `testTonyResourcesFlag` test? Seems like there will no longer be a `test.zip`, only a `test20.zip`.

oliverhu: yah.. that is the reason why tests are failling. Fixing

tony-core/src/main/java/com/linkedin/tony/LocalizableResource.java
oliverhu: agreed, this is confusing, will do some housekeeping for this class

erwa: Thanks. The new name `completeResourceString` and the Javadoc make it much clearer.

PR #258, opened 2019-04-25
Bump version and update shadeJar to exclude guava dependencies

oliverhu:
Existing com.google.guava in fat jar conflicts with Hadoop 3.3

PR #259, opened 2019-04-25
Bump version in preparation for publishing new TonY version with metrics to Maven

erwa:
This is so that Dr. Elephant can depend on TonY and the metric-parsing classes in TonY.


oliverhu:
i am bumping in this . https://github.com/linkedin/TonY/pull/258

erwa:
Oh, didn't see that. Will close this PR then.

Issue #262, opened 2019-05-01
TonY Portal synchronously scans and moves folders from intermediate to finished every time the main page is loaded


erwa:
Currently, every time the main page (JobsMetadataPage) is loaded, TonY Portal synchronously scans the `intermediate` directory and moves all finished applications to the `finished` directory. This can significantly delay load time. The scanning and moving should be done in a periodic background thread.

PR #263, opened 2019-05-02
Fix #261, #262: Refactor moving intermediate to finished to separate thread, create yyyy/mm/dd dirs based on finished time embedded in jhist file instead of directory modification time

### LINKED ISSUE #261: TonY Portal creates finished dir based off start time rather than finish time
Currently, TonY Portal uses the modification time of the `intermediate/<appId>` directory to create the `finished/yyyy/mm/dd` directory. The modification time is approximately the start time, but we should be using the finish time, which is embedded in the `intermediate/<appId>/<appId>-<start>-<end>-<user>-<status>.jhist` file name.

erwa:
* Added unit test
* Tested locally and on our cluster

To provide more context, previously, the moving of the job directories from the `intermediate` directory to the `finished` directory was being done in `JobsMetadataPageController` every time it was called (whenever the main TonY Portal page was loaded). (This is issue #262.) I moved this code out into a background thread (`HistoryFileMover`).

Also, previously, the `finished` directory folder creation code (which creates `yyyy/mm/dd`) was basing the date off the last modification time of the job directory in `intermediate`, rather than the finish time embedded in the jhist file name itself (`<appId>-<startTime>-<ENDTIME>-<user>-<status>.jhist`). The last modification time is basically when the job started, so it might be several days before the endtime (this was issue #261). I also updated the folder creation code to use the `ENDTIME` embedded in the jhist file name instead.

### DIFF
--- a/tony-portal/app/history/HistoryFileMover.java
+++ b/tony-portal/app/history/HistoryFileMover.java
@@ -69,7 +69,7 @@ public class HistoryFileMover {
         // to compensate.
         path.append(Path.SEPARATOR).append(source.getName());
       }
-      Utils.createDir(fs, new Path(path.toString()), Constants.PERM770);
+      Utils.createDirIfNotExists(fs, new Path(path.toString()), Constants.PERM770);
 
       Path dest = new Path(path.toString());
       LOG.info("Moving " + source + " to " + dest);
--- a/tony-portal/app/history/HistoryFileMover.java
+++ b/tony-portal/app/history/HistoryFileMover.java
@@ -1,0 +1,85 @@
+package history;
+
+import com.linkedin.tony.Constants;
+import com.linkedin.tony.TonyConfigurationKeys;
+import com.linkedin.tony.util.ParserUtils;
+import com.linkedin.tony.util.Utils;
+import com.typesafe.config.Config;
+import hadoop.Requirements;
+import java.io.IOException;
+import java.util.Date;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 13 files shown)


tony-portal/app/history/HistoryFileMover.java
```diff
@@ -0,0 +1,87 @@
...
+  private final Path finishedDir;
+
+  @Inject
+  public HistoryFileMover(Config appConf, Requirements requirements) {
+    fs = requirements.getFileSystem();
+    intermediateDir = requirements.getIntermDir();
+    finishedDir = requirements.getFinishedDir();
+
+    ScheduledExecutorService scheduledThreadPool = Executors.newScheduledThreadPool(1);
+    long moverIntervalMs = ConfigUtils.fetchIntConfigIfExists(appConf,
```
hungj: fetchLongConfigIfExists?

tony-portal/app/history/HistoryFileMover.java
```diff
@@ -0,0 +1,87 @@
...
+      Path source = new Path(jhistFilePath).getParent();
+      StringBuilder path = new StringBuilder(finishedDir.toString());
+      Date endDate = new Date(ParserUtils.getCompletedTimeFromJhistFileName(jhistFilePath));
+      path.append(Path.SEPARATOR).append(ParserUtils.getYearMonthDayDirectory(endDate));
+      if (fs.getScheme().equals("file")) {
+        // On Mac, local filesystem will copy contents of source dir to dest dir, so we have to append the dir name
+        // to compensate.
+        path.append(Path.SEPARATOR).append(source.getName());
+      }
+      Utils.createDir(fs, new Path(path.toString()), Constants.PERM770);
```
hungj: should we rename this to createDirIfNotExists?

tony-portal/app/history/HistoryFileMover.java
erwa: The method actually returns an `int` and then I just rely on Java's auto conversion to `long`.

tony-portal/app/history/HistoryFileMover.java
erwa: Good idea.

Issue #271, opened 2019-05-03
Config file written to history location should be named tony-final.xml rather than config.xml


erwa:
The final config file we ship to the AM and TaskExecutors is named `tony-final.xml`. We should name the config file written to the history location the same thing rather than renaming it to `config.xml`.

PR #272, opened 2019-05-03
Fix #270, #271: Move APP_TYPE to Constants, change history config file name to tony-final.xml

### LINKED ISSUE #270: Move TonyClient.APP_TYPE to Constants and make public
Currently, `TonyClient` defines `private static final String APP_TYPE = "TONY"`. We should move this to `Constants` and make it `public` so that it can be referenced in other applications like Dr. Elephant.

erwa:
* Moved `TonyClient.APP_TYPE` to `Constants` (#270). I would like to reference it from Dr. Elephant as part of https://github.com/linkedin/dr-elephant/issues/553.
* Changed the name of the config file written to the history location from `config.xml` to `tony-final.xml` (#271)


erwa:
Pushed new commit to fix broken test.

PR #283, opened 2019-05-09
Fix #268 Expose Azkaban flow information in YARN application tag.

### LINKED ISSUE #268: Expose Azkaban flow information in YARN application tag
The Azkaban flow information is useful in debugging settings to tracking down which Azkaban flow a Tony job belongs to. We could expose that information via YARN application tag for that purpose.

haibchen:
1) Extract flow info from Job Properties provided by Azkaban
2) Pass the flow info as YARN application tags in configuration

### DIFF
--- a/tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
+++ b/tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
@@ -4,15 +4,22 @@
  */
 package com.linkedin.tony.azkaban;
 
+import azkaban.flow.CommonJobProperties;
+import azkaban.utils.FileIOUtils;
+import azkaban.utils.Props;
+
 import com.google.inject.AbstractModule;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
 
-import azkaban.utils.FileIOUtils;
-import azkaban.utils.Props;
 import java.io.File;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.log4j.Logger;
 import org.testng.Assert;
 import org.testng.annotations.BeforeTest;
@@ -57,10 +64,44 @@ public class TestTonyJob {
       }
     };
     String args = tonyJob.getMainArguments();
-    Assert.assertTrue(new File(tonyJob.getWorkingDirectory(), "_tony-conf-test_tony_job/tony.xml").exists());
     Assert.assertTrue(args.contains(TonyJobArg.HDFS_CLASSPATH.tonyParamName + " hdfs://nn:8020"));
     Assert.assertTrue(args.contains(TonyJobArg.SHELL_ENV.tonyParamName + " E2=e2"));
     Assert.assertTrue(args.contains(TonyJobArg.SHELL_ENV.tonyParamName + " E1=e1"));
+  }
+
+  /**
+   * Check if the flow level information is passed to the tony job through configuration.
+   */
+  @Test
+  public void testFlowInfoPropagation() {
+    final Props jobProps = new Props();
+    jobProps.put(TonyJobArg.HDFS_CLASSPATH.azPropName, "hdfs://nn:8020");
+    jobProps.put(CommonJobProperties.PROJECT_NAME, "unit_test");
+    jobProps.put(CommonJobProperties.FLOW_ID, "1");
+    jobProps.put(CommonJobProperties.EXEC_ID, "0");
+
+    final TonyJob tonyJob = new TonyJob("test_tony_job", new Props(), jobProps, log) {
+      @Override
+      public String getWorkingDirectory() {
+        return System.getProperty("java.io.tmpdir");
+      }
+    };
+
+    Configuration conf = tonyJob.getTonyJobConf();
+    Set<String> values = new HashSet<>(
+        conf.getStringCollection(TonyJob.TONY_APPLICATION_TAGS));
+    values.remove("");
+
+    Map<String, String> parsedTags = new HashMap<>();
+    for (String value : values) {
+      String[] pair = value.split(":");
+      Assert.assertTrue(pair.length == 2);
+      parsedTags.put(pair[0], pair[1]);
+    }
+
+    Assert.assertEquals(parsedTags.get(CommonJobProperties.EXEC_ID), "0");
+    Assert.assertEquals(parsedTags.get(CommonJobProperties.FLOW_ID), "1");
+    Assert.assertEquals(parsedTags.get(CommonJobProperties.PROJECT_NAME), "unit_test");
   }
 
   @Test
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -114,6 +114,7 @@ public class TonyClient implements AutoCloseable {
   private String pythonBinaryPath = null;
   private String pythonVenv = null;
   private String srcDir = null;
+  private Set<String> applicationTags;
   private String hdfsClasspath = null;
   private String executes;
   private long appTimeout;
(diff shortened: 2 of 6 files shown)


erwa:
CI had failing test:
```
Gradle suite > Gradle test > com.linkedin.tony.azkaban.TestTonyJob.testFlowInfoPropagation FAILED
    java.lang.AssertionError: expected object to not be null
        at org.testng.Assert.fail(Assert.java:89)
        at org.testng.Assert.assertNotNull(Assert.java:399)
        at org.testng.Assert.assertNotNull(Assert.java:384)
        at com.linkedin.tony.azkaban.TestTonyJob.testFlowInfoPropagation(TestTonyJob.java:98)
```

https://travis-ci.org/linkedin/TonY/builds/530500852?utm_source=github_status&utm_medium=notification

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
```diff
@@ -63,6 +70,48 @@ public String getWorkingDirectory() {
...
+
+    final TonyJob tonyJob = new TonyJob("test_tony_job", new Props(), jobProps, log) {
+      @Override
+      public String getWorkingDirectory() {
+        return System.getProperty("java.io.tmpdir");
+      }
+    };
+    tonyJob.getMainArguments();
+
+    org.apache.hadoop.conf.Configuration conf =
```
erwa: Add a `import org.apache.hadoop.conf.Configuration` rather than using the full class name here.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -380,6 +397,10 @@ public boolean init(String[] args) throws ParseException, IOException {
     // src_dir & hdfs_classpath flags are for compatibility.
     srcDir = cliParser.getOptionValue("src_dir");
 
+    applicationTags = convertStringToValueSet(
```
erwa: Rather than adding a `convertStringToValueSet`, you could also just use
```
new HashSet(tonyConf.getStringCollection(...);
```

tony-core/src/test/java/com/linkedin/tony/TestTonyClient.java
```diff
@@ -104,4 +104,16 @@ public void testValidateTonyConfRequestedGpusUnderLimit() {
...
+  @Test
+  public void testTonyApplciatonTags() throws Exception {
+    final String applicationTags = "flow_id:1,exec_id:0,proj_name:demo";
+    String[] args = { "-conf",
+        TonyConfigurationKeys.APPLICATION_TAGS + "=" + applicationTags };
+    TonyClient client = new TonyClient();
+    client.init(args);
+    assertEquals(
+        applicationTags,
+        client.getTonyConf().get(TonyConfigurationKeys.APPLICATION_TAGS));
```
erwa: Nit: based on the Javadoc of TestNG's `assertEquals`, the `actual` argument should come first, then the `expected`. So I would swap lines 116 and 117.

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
```diff
@@ -63,6 +70,48 @@ public String getWorkingDirectory() {
...
+        new org.apache.hadoop.conf.Configuration(false);
+    conf.addResource(
+        new Path(tonyJob.getWorkingDirectory() + "_tony-conf-test_tony_job/tony.xml"));
```
erwa: I think it's not good that `TonyJob.getMainArguments()` has the side effect of writing out this XML file. Can you move the XML file writing logic out into a separate method that's called from the constructor?

BTW, I don't think it's necessary to write a test verifying that the conf was written to the `tony.xml`; I think it's enough to verify that the application tags are added to the `Configuration` object. (You might need to add a method to `TonyJob` that exposes the `Configuration` for testing purposes.)

tony-core/src/test/java/com/linkedin/tony/TestTonyClient.java
```diff
@@ -104,4 +104,16 @@ public void testValidateTonyConfRequestedGpusUnderLimit() {
     conf.setInt(TonyConfigurationKeys.getMaxTotalResourceKey(Constants.GPUS), 10);
     TonyClient.validateTonyConf(conf);
   }
+
+  @Test
+  public void testTonyApplciatonTags() throws Exception {
```
erwa: Applciaton -> Application

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
```diff
@@ -63,6 +70,48 @@ public String getWorkingDirectory() {
...
+
+    String applicationTags = conf.get(TonyJob.TONY_APPLICATION_TAGS);
+    Assert.assertNotNull(applicationTags);
+
+    Set<String> values = new LinkedHashSet<>(
```
erwa: Why a _Linked_HashSet as opposed to an ordinary `HashSet`? Also, instead of doing this parsing yourself, you can just use `conf.getStringCollection()`, which returns a `Collection<String>`.

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
```diff
@@ -4,15 +4,22 @@
  */
 package com.linkedin.tony.azkaban;
 
+import azkaban.flow.CommonJobProperties;
 import com.google.inject.AbstractModule;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
 
 import azkaban.utils.FileIOUtils;
 import azkaban.utils.Props;
```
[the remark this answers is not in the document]
haibchen: Fixed

tony-core/src/test/java/com/linkedin/tony/TestTonyClient.java
haibchen: Good catch

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
haibchen: Fixed

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
haibchen: Addressed

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
haibchen: That's a good point. I split the processing Conf and writing it to xml file into two separate methods,  getJobConfiguration() and  setupJobConfigurationFile(). The former is called from the TonyJob constructor, the latter does IO, which I don't think is suitable to put in the constructor. Instead, it is called in run() method as a prep step as setupHadoopOpts().

tony-core/src/test/java/com/linkedin/tony/TestTonyClient.java
haibchen: addressed

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
haibchen: Addressed

tony-azkaban/src/test/java/com/linkedin/tony/azkaban/TestTonyJob.java
```diff
@@ -63,6 +70,41 @@ public String getWorkingDirectory() {
...
+      @Override
+      public String getWorkingDirectory() {
+        return System.getProperty("java.io.tmpdir");
+      }
+    };
+
+    Configuration conf = tonyJob.getTonyJobConf();
+    Set<String> values = new HashSet<>(
+        conf.getStringCollection(TonyJob.TONY_APPLICATION_TAGS));
+    values.remove("");
```
erwa: I don't think this is necessary. The default value of `TONY_APPLICATION_TAGS` is `""` (empty string), which `Configuration` interprets as `null`, and `getStringCollection()` should return an empty collection.

PR #285, opened 2019-05-15
Fix #284. Token for application rpc should support IP service name.

yuanzac:

PR #298, opened 2019-05-24
Apply #285 in master: Token for application rpc should support IP ser…

oliverhu:
…vice name

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jdk:
   - oraclejdk8
 env:
   - HADOOP_VERSION=3.1.1
-script: "./gradlew check --stacktrace --info"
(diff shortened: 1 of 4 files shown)


oliverhu:
Try to fix the build issue in #285

oliverhu:
Difference between this & https://github.com/linkedin/TonY/pull/285 is I changed the default of mini cluster to:
```yarnConf.setBoolean(CommonConfigurationKeys.HADOOP_SECURITY_TOKEN_SERVICE_USE_IP, false);```..

oliverhu:
I don't understand either.. that was the initial error I saw when I checked out Zac's patch, after rebasing master, the error is gone from local but still fails in Travis :(

PR #300, opened 2019-05-24
Fix broken docker support

oliverhu:
There is a typo during reflection of using docker runtime.


oliverhu:
Yah, I am not sure how to unit test docker environment w/ MiniYARN yet, need some further investigation, logged #304 to follow up

PR #301, opened 2019-05-24
Fix #98: TonY Portal should enforce retention on history files

### LINKED ISSUE #98: TonY Portal should enforce retention on history files

erwa:
* Added HistoryFileCleaner thread
* Added configs `tony.history.retention-sec` (default 30 days) and `tony.history.cleaner-interval-ms` (default 6 hours)
* Added unit tests
* Minor tweaks to startTonyPortal.sh
* Tested in real cluster

### DIFF
--- a/tony-core/src/main/resources/tony-default.xml
+++ b/tony-core/src/main/resources/tony-default.xml
@@ -264,13 +264,13 @@
   <property>
     <name>tony.history.retention-sec</name>
     <value>2592000</value>
-    <description>How many seconds to retain history files for.</description>
+    <description>How many seconds to retain history files for. Default is 30 days.</description>
   </property>
 
   <property>
-    <name>tony.history.cleaner-interval-ms</name>
+    <name>tony.history.purger-interval-ms</name>
     <value>21600000</value>
-    <description>Frequency at which to run the cleaner thread.</description>
+    <description>Frequency at which to run the purger thread. Default is 6 hours.</description>
   </property>
 
   <property>
--- a/tony-core/src/main/resources/tony-default.xml
+++ b/tony-core/src/main/resources/tony-default.xml
@@ -262,6 +262,18 @@
   </property>
 
   <property>
+    <name>tony.history.retention-sec</name>
+    <value>2592000</value>
+    <description>How many seconds to retain history files for.</description>
+  </property>
+
+  <property>
+    <name>tony.history.cleaner-interval-ms</name>
+    <value>21600000</value>
+    <description>Frequency at which to run the cleaner thread.</description>
+  </property>
+
+  <property>
     <description>
       Maximum (approximate) number of cache entries TonY Portal
       will store for each cache (config, metadata, event)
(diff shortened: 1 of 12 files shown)


tony-core/src/main/resources/tony-default.xml
```diff
@@ -261,6 +261,18 @@
     finished directory.</description>
   </property>
 
+  <property>
+    <name>tony.history.retention-sec</name>
+    <value>2592000</value>
+    <description>How many seconds to retain history files for.</description>
```
oliverhu: can you put how many days is 2592000 in the desc? same below

tony-portal/app/history/HistoryFileCleaner.java
```diff
@@ -0,0 +1,110 @@
...
+import javax.inject.Singleton;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import play.Logger;
+import utils.ConfigUtils;
+
+
+@Singleton
+public class HistoryFileCleaner {
```
oliverhu: HistoryFilePurger sounds more professional ?

tony-portal/app/history/HistoryFileCleaner.java
```diff
@@ -0,0 +1,110 @@
...
+        LOG.error("Encountered exception while cleaning history directories", e);
+      }
+    }, 0, cleanerIntervalMs, TimeUnit.MILLISECONDS);
+  }
+
+  /**
+   * Deletes all year/month/day directories in the finished dir prior to the cutoff date.
+   */
+  @VisibleForTesting
+  static void cleanFinishedDir(FileSystem fs, Path finishedDir, LocalDate cutOffDate) throws IOException {
```
oliverhu: why is cleanFinishedDir so different from cleanIntermediateDir ?

tony-portal/app/history/HistoryFileCleaner.java
erwa: Agreed.

tony-portal/app/history/HistoryFileCleaner.java
erwa: The structure of finished is yyyy/mm/dd/<app_id> whereas intermediate directly contains <app_id> folders.

For purging finished, we have to parse yyyy/mm/dd and interpret as a date, whereas for intermediate, I'm just using the last modification time of the <app_id> directory as a heuristic.

tony-core/src/main/resources/tony-default.xml
erwa: Good idea. Updated.

Issue #318, opened 2019-06-06
NullPointerException in ParserUtils.parseConfig if config is missing <final> element


erwa:
Currently, ParserUtils assumes that each `<property>` in the job config has the following 4 subelements:
* `<name>`
* `<value>`
* `<final>`
* `<source>`

`<final>` may not always be present, causing a NPE:

```
2019-06-06 00:13:10,516 [ERROR] from history.HistoryFileMover in pool-4-thread-1 - Encountered exception while moving history directories
java.lang.NullPointerException: null
        at com.linkedin.tony.util.ParserUtils.parseConfig(ParserUtils.java:214)
        at cache.CacheWrapper.updateCaches(CacheWrapper.java:90)
        at history.HistoryFileMover.moveIntermediateToFinished(HistoryFileMover.java:72)
        at history.HistoryFileMover.lambda$new$0(HistoryFileMover.java:62)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

To be more robust, we should add null checks.

PR #320, opened 2019-06-06
Bump version in preparation for publishing to Maven Central #318

erwa:


oliverhu:
We could automate version bump. What we did before was to write the version to a file and pick up the version in gradle.

erwa:
Yeah, that would be ideal. The version bump needs to be part of an existing PR though, since PRs cannot get auto-merged (they need a human approval).

PR #334, opened 2019-06-23
Support different docker images for different job types

oliverhu:
Also fixed that the images were not properly set for non AM containers

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -1071,6 +1071,8 @@ public class ApplicationMaster {
       LOG.info("Setting Container [" + container.getId() + "] for task [" + task.getId() + "]..");
 
       Map<String, String> containerLaunchEnv = new ConcurrentHashMap<>(containerEnv);
+      Utils.parseDockerConfigs(tonyConf, containerEnv);
+
       /*
        * Add additional environment vars. We always set job_name task_index & task_num and
        * task_num and TaskExecutor is responsible for setting up the actual shell environment
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -1079,7 +1079,7 @@ public class ApplicationMaster {
        */
       String jobName = task.getJobName();
       String taskIndex = task.getTaskIndex();
-      Map<String, String> dockerEnv = Utils.getDockerContainerEnv(tonyConf, jobName);
+      Map<String, String> dockerEnv = Utils.getContainerEnvForDocker(tonyConf, jobName);
       containerLaunchEnv.putAll(dockerEnv);
       containerLaunchEnv.put(Constants.JOB_NAME, jobName);
       containerLaunchEnv.put(Constants.TASK_INDEX, taskIndex);
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -1071,7 +1071,6 @@ public class ApplicationMaster {
       LOG.info("Setting Container [" + container.getId() + "] for task [" + task.getId() + "]..");
 
       Map<String, String> containerLaunchEnv = new ConcurrentHashMap<>(containerEnv);
-      Utils.parseDockerConfigs(tonyConf, containerLaunchEnv);
 
       /*
        * Add additional environment vars. We always set job_name task_index & task_num and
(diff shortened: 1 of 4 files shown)


tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -618,5 +618,50 @@ public static void extractResources() {
     }
   }
 
+  /**
+   * Parses Docker related configs and sets the appropriate container environment variables if Docker is available.
+   * Uses reflection to support older versions of Hadoop.
+   */
+  public static boolean parseDockerConfigs(Configuration tonyConf, Map<String, String> containerEnv, String jobType) {
```
erwa: Please explain in the Javadoc what the boolean returned means.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -618,5 +618,50 @@ public static void extractResources() {
...
+            String envDockerImage = (String) dockerRuntimeClass.getField(Constants.ENV_DOCKER_CONTAINER_IMAGE).get(null);
+            containerEnv.put(envContainerType, "docker");
+            containerEnv.put(envDockerImage, imagePath);
+          } catch (NoSuchFieldException e) {
+            LOG.error("Field " + Constants.ENV_CONTAINER_TYPE + " or " + Constants.ENV_DOCKER_CONTAINER_IMAGE + " not "
+                          + "found in " + containerRuntimeClass.getName());
+            return false;
+          } catch (IllegalAccessException e) {
+            LOG.error("Unable to access " + Constants.ENV_CONTAINER_TYPE + " or "
+                          + Constants.ENV_DOCKER_CONTAINER_IMAGE + " fields.");
```
erwa: Please include `, e` (log the exception)

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -618,5 +618,50 @@ public static void extractResources() {
...
+        }
+        if (dockerRuntimeClass != null) {
+          try {
+            String envContainerType = (String) containerRuntimeClass.getField(Constants.ENV_CONTAINER_TYPE).get(null);
+            String envDockerImage = (String) dockerRuntimeClass.getField(Constants.ENV_DOCKER_CONTAINER_IMAGE).get(null);
+            containerEnv.put(envContainerType, "docker");
+            containerEnv.put(envDockerImage, imagePath);
+          } catch (NoSuchFieldException e) {
+            LOG.error("Field " + Constants.ENV_CONTAINER_TYPE + " or " + Constants.ENV_DOCKER_CONTAINER_IMAGE + " not "
+                          + "found in " + containerRuntimeClass.getName());
```
erwa: Include `e`

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
erwa: Can we also rename this method something like `setDockerEnvironmentVariables`? `parseXXX` doesn't sound like it has side effects, but this method does have side effects.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -618,5 +618,50 @@ public static void extractResources() {
...
 
+  /**
+   * Parses Docker related configs and sets the appropriate container environment variables if Docker is available.
+   * Uses reflection to support older versions of Hadoop.
+   */
+  public static boolean parseDockerConfigs(Configuration tonyConf, Map<String, String> containerEnv, String jobType) {
+    if (tonyConf.getBoolean(TonyConfigurationKeys.DOCKER_ENABLED, TonyConfigurationKeys.DEFAULT_DOCKER_ENABLED)) {
+      String imagePath = tonyConf.get(TonyConfigurationKeys.getContainerDockerKey());
+      if (tonyConf.get(TonyConfigurationKeys.getDockerImageKey(jobType)) != null) {
+        imagePath = tonyConf.get(TonyConfigurationKeys.getDockerImageKey(jobType));
```
erwa: Use local variable for `tonyConf.get(TonyConfigurationKeys.getDockerImageKey(jobType))` instead of repeating.

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -1071,13 +1071,15 @@ public void run() {
...
       Map<String, String> containerLaunchEnv = new ConcurrentHashMap<>(containerEnv);
+
       /*
        * Add additional environment vars. We always set job_name task_index & task_num and
        * task_num and TaskExecutor is responsible for setting up the actual shell environment
        * for different deep learning frameworks.
        */
       String jobName = task.getJobName();
       String taskIndex = task.getTaskIndex();
+      Utils.parseDockerConfigs(tonyConf, containerLaunchEnv, jobName);
```
erwa: Are you going to check the boolean return value?

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
oliverhu: Yah, that makes sense. I'll update the signature of this function

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
oliverhu: that is too much stack trace. I think our error info is enough

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
oliverhu: same here

PR #335, opened 2019-06-25
Close TonyClient on Cluster Launcher

pdambrauskas:
TonyClient was never closed, so config files & penv.zip were never deleted from resources directory


oliverhu:
Thanks for the contribution!

Issue #82, opened 2018-11-14
build TonY on CDH 5.15.1(Hadoop 2.6.0)


iamabug:
Is there some feature that is strictly not supported on Hadoop of lower version ? I try to build TonY on Hadoop 2.6.0 and get the following:

```bash
$ ./gradlew build -x test

> Task :tony-core:compileJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: /home/xxx/TonY/tony-core/src/main/java/com/linkedin/tony/rpc/impl/ApplicationRpcClient.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
Unknown file extension: tony-core/src/main/resources/META-INF/services/org.apache.hadoop.security.SecurityInfo
Unknown file extension: tony-core/src/test/resources/test.tar
Unknown file extension: tony-core/src/test/resources/test.tar.gz
Unknown file extension: tony-core/src/test/resources/test.zip

> Task :tony-history-server:compilePlayBinaryScala
Pruning sources from previous analysis, due to incompatible CompileSetup.

> Task :tony-history-server:compilePlayBinaryTests
Pruning sources from previous analysis, due to incompatible CompileSetup.
Note: /home/xxx/TonY/tony-history-server/test/utils/TestHdfsUtils.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :tony-history-server:testPlayBinary

controllers.BrowserTest > test FAILED
    java.lang.RuntimeException
        Caused by: akka.stream.impl.io.ConnectionSourceStage$$anon$2$$anon$1
            Caused by: java.net.BindException

12 tests completed, 1 failed

> Task :tony-history-server:testPlayBinary FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':tony-history-server:testPlayBinary'.
> There were failing tests. See the report at: file:///home/xxx/TonY/tony-history-server/build/playBinary/reports/test/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 5.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/4.10.2/userguide/command_line_interface.html#sec:command_line_warnings

[...]

PR #336, opened 2019-06-25
Pass container ApplicationAttemptId to security handling

pdambrauskas:
We are using TonY on `2.6.0-cdh5.14.2` hadoop distribution.
`ClientToAMTokenIdentifier` has a `write` method, which is triggered by `SecretManager`:
```java
  @Override
  public void write(DataOutput out) throws IOException {
    out.writeLong(this.applicationAttemptId.getApplicationId()
      .getClusterTimestamp());
    out.writeInt(this.applicationAttemptId.getApplicationId().getId());
    out.writeInt(this.applicationAttemptId.getAttemptId());
    this.clientName.write(out);
  }
```
as you can see it calls `(this.applicationAttemptId.getApplicationId().getClusterTimestamp()`, which results in NullPointerException.

This problem is not present on hadoop 2.7.3, (or non-cdh version of hadoop 2.6.0) since write method is using protobuf:
```java
  @Override
  public void write(DataOutput out) throws IOException {
    out.write(proto.toByteArray());
  }
```

Not sure if this change is worth of being merged but maybe it would make things easier for peaple, trying to build TonY against older versions of CDH.

Also it looks somewhat wrong to pass null instead attemptId, since we do have it present on container.

Related: https://github.com/linkedin/TonY/issues/82#issuecomment-504351334

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -98,7 +98,6 @@ public class ApplicationMaster {
   /**
    * Metadata + History Server related variables
    */
-  private ApplicationAttemptId appAttemptID = null;
   private String appIdString;
   private FileSystem resourceFs; // FileSystem used to access resources for the job, like jars and zips
   private FileSystem historyFs;  // FileSystem used to write history-related files like config and events.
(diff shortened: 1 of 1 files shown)


oliverhu:
looks like a small change, CI failed for a dependency reason, I'm restarting the build.

pdambrauskas:
> Could not GET 'https://repo.maven.apache.org/maven2/com/linkedin/azkaban/azkaban-common/3.58.0/azkaban-common-3.58.0.pom'. Received status code 403 from server: Forbidden

Doesn't seem like a problem caused by this change :thinking:

erwa:
Agreed. Seems like a transient issue. I didn't hit this in my test PR: https://github.com/linkedin/TonY/pull/337 --> https://travis-ci.org/linkedin/TonY/builds/550527949.

I've kicked off another run of Travis for your change here: https://travis-ci.org/linkedin/TonY/builds/550441747

Issue #342, opened 2019-07-08
TonY can't get delegation token when using ViewFS


ChengbingLiu:
In `TonyClient`, `fs.getDelegationToken` is used to get delegation token, which is an internal method and should not be called according to Javadoc. Current implementation can not handle ViewFS. We should use `fs.addDelegationTokens` instead.

PR #343, opened 2019-07-08
Use addDelegationTokens API to support ViewFS (#342)

ChengbingLiu:


tony-core/src/main/java/com/linkedin/tony/security/TokenCache.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private TokenCache() {
+    // Prevent instantiation
+  }
+
+  /**
+   * Convenience method to obtain delegation tokens from namenodes corresponding
+   * to the paths passed.
+   * 
+   * @param credentials
```
oliverhu: could you add some doc for the parameters? similar to the function below

oliverhu:
lgtm, we need to validate it first

tony-core/src/main/java/com/linkedin/tony/security/TokenCache.java
ChengbingLiu: I just added comments and fix method declaration for `obtainTokensForNamenodesInternal`

tony-core/src/main/java/com/linkedin/tony/security/TokenCache.java
```diff
@@ -0,0 +1,74 @@
+/**
+ * Copyright 2018 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
```
erwa: Nit: 2018 -> 2019

tony-core/src/main/java/com/linkedin/tony/security/TokenCache.java
```diff
@@ -0,0 +1,74 @@
...
+   * 
+   * @param credentials
+   *          cache in which to add new delegation tokens
+   * @param ps
+   *          array of paths
+   * @param conf
+   *          configuration
+   * @throws IOException
+   */
+  public static void obtainTokensForNamenodes(Credentials credentials, Path[] ps, Configuration conf, String renewer) throws IOException {
```
erwa: Can you change `ps` to `paths` to avoid confusion with parameter server (which is also commonly abbreviated PS)?

erwa:
Tested on our cluster. Works fine. @ChengbingLiu, can you address my comments? I will commit after that.

ChengbingLiu:
@erwa Code updated, please review again.

erwa:
Thanks for addressing my comments, @ChengbingLiu. Will merge once tests pass.

Issue #344, opened 2019-07-08
Exclude evaluator from cluster spec and include it in untracked.jobtypes


ChengbingLiu:
To fully support evaluator, two changes should be made:
1. evaluator should not be included in cluster spec 
2. evaluator should better be included in `untracked.jobtypes` by default, along with ps

erwa:
Hi @ChengbingLiu,

I believe it should not cause problems to include evaluator in the cluster_spec. TF seems to just ignore keys it doesn't recognize. We want TonY to be flexible and allow users to launch auxiliary task types (e.g.: to host sidecar tables) and have them all included in the cluster_spec, which is a central place all tasks can be parsed from.

As discussed in #345, in general you *do* want to wait for the evaluator to finish (to do a final evaluation after the workers have finished training). You should specify a max steps that is used by both the workers and evaluator to ensure the evaluator terminates. We have seen issues where user specified both num_epochs and max_steps, and num_epochs was reached before max_steps, and evaluator just waited forever. By only relying on max_steps, you can avoid this issue.

Of course, for your use case, if you want, you can always add evaluator to the untracked.jobtypes in your job config.

PR #345, opened 2019-07-08
Exclude evaluator from cluster spec and include it in untracked.jobtypes (#344)

ChengbingLiu:


oliverhu:
Thanks for the PR! From our experience, we do want to wait for evaluator to exit before ending the training, otherwise the last evaluation would miss. For cluster spec, we are using it as a generic environment to map all job types to their hosts, is there a reason to exclude evaluator from there?

ChengbingLiu:
@oliverhu I wonder if you use estimator API. We discover that the evaluator of estimator API will not exit even when all workers and chief have finished training and exited. It is always waiting for new checkpoints to arrive and do the evaluation.

The reason evaluator should not be in cluster spec is in the official docs below:
https://www.tensorflow.org/api_docs/python/tf/estimator/train_and_evaluate
"Evaluator is a special task that is not part of the training cluster." Previously with TonY, the cluster always wait for the "evaluator task" to start up. However the evaluator thinks itself not part of the cluster, so it does not register itself to all the workers, therefore the training process can never begin.

oliverhu:
Yes, we use estimator and observed the same issue when specified num of epochs does not have enough max_steps. remove epochs would make evaluator stop.

ChengbingLiu:
Setting max_steps is a solution. However, it's not easy to get the result of last evaluation performed by evaluator. That's why we add a separate evaluation process after the training job in production to judge if the saved model is above some standard.

As for cluster spec, what do you think about it?

oliverhu:
i see, as this is configuration, I'd like to the default.

for the cluster spec, I don't want to make it TensorFlow only (when we execute PyTorch training, we could also derive configurations from the same cluster spec), so I'm not intending to exclude other job types from cluster spec, is this additonal field blocking your TensorFlow training?

ChengbingLiu:
> is this additonal field blocking your TensorFlow training?
Yes, the evaluator thinks itself not a part of the cluster, so it does not register itself to all the workers, therefore the training process can never begin.

> I don't want to make it TensorFlow only (when we execute PyTorch training, we could also derive configurations from the same cluster spec)
ClusterSpec is handled in com.linkedin.tony.tensorflow.TonySession.java, which I think is TensorFlow-specific. Isn't it?

erwa:
@ChengbingLiu: 

> That's why we add a separate evaluation process after the training job in production to judge if the saved model is above some standard.

It sounds like you run one TonY job with only workers and ps. Then you run another job that runs evaluation. But you also said

> Yes, the evaluator thinks itself not a part of the cluster, so it does not register itself to all the workers, therefore the training process can never begin.

If you run training and evaluation in separate jobs, why do you have "evaluator" tasks in your TonY job?

In any case, if you run evaluator tasks as part of TonY job, the Tony TaskExecutor should automatically register them and add them to the cluster spec.

Can you elaborate on what your problem is? I'm still not very clear about it. Also see my comments in #344.

ChengbingLiu:
@erwa Sorry about confusion, I will try to fully describe the way we use TensorFlow and the problem we met.

Our whole workflow contains three parts:
- training with estimator API (chief/worker/ps/evaluator)
- after-training evaluation
- deploy the model with TF serving.

In the first part, the evaluator performs evaluation every 10-20 minutes, so that we can observe test auc during the training process. I will certainly try your method of setting max_steps to get the final test auc.

But I'm not sure how to get the auc value by program so that I could judge whether the model is qualified for production. The second part mainly solves this problem. We start a separate evaluator (which is not launched by TonY for now) to get the final test auc.

If the test auc is below some threshold, the workflow will exit, so that the unqualified model will not be deployed. Otherwise it will be deployed with TF serving.


Now my main point is that evaluator should not be in cluster spec defined in TF_CONFIG for two reasons:
1. If evaluator is included in cluster spec, the training will never start. We experienced that with TonY 0.3.8. Maybe it's due to a mistake in program or something else. So maybe we can just ignore this one.
2. The official doc says "evaluator is not part of training cluster". See https://www.tensorflow.org/api_docs/python/tf/estimator/RunConfig . I think TonY should conform to that.

oliverhu:
We don't see this issue with evaluator address in TF_CONFIG using estimator in our training jobs, let me double check with him and pull him here.

zhangxuhong:
@ChengbingLiu Thanks for the contribution. Yes, i agree we should remove `evaluator` from cluster spec. However, in our internal usecase, I did not see having `evaluator` in cluster spec is preventing the training job to start. This is confirmed both in 1.12 and 1.13.

As for the max_steps, this is the recommended way by the official doc. I don't have experience of getting the last eval value, but i think you might be able to get it from the checkpoints after the `train_and_evaluate` call:
> Stop condition: In order to support both distributed and non-distributed configuration reliably, the only supported stop condition for model training is train_spec.max_steps. If train_spec.max_steps is None, the model is trained forever. Use with care if model stop condition is different. For example, assume that the model is expected to be trained with one epoch of training data, and the training input_fn is configured to throw OutOfRangeError after going through one epoch, which stops the Estimator.train. For a three-training-worker distributed configuration, each training worker is likely to go through the whole epoch independently. So, the model will be trained with three epochs of training data instead of one epoch.

pfxuan:
@zhangxuhong same here. The `evaluator` seems working fine with the `train_and_evaluate` call on our cluster.

ChengbingLiu:
@oliverhu @zhangxuhong @pfxuan Yes, you are right. I validated on latest code and it worked fine.

But according to official doc, evaluator is not part of cluster or cluster spec. Shouldn't TonY conform to it?

oliverhu:
As I said before, we intentionally want to decouple the cluster spec from serving TensorFlow only. For example, we could use the same spec to parse cluster information and set up a PyTorch training.

ChengbingLiu:
Got your point. Closing this PR.

PR #353, opened 2019-07-12
Allow sorting

StanfordMCP:
This change updates the TonY portal page to provide a sorted table of TonY process records, starting from most recent completion date.  This change also provides a toString() method to the JobMetadata class to print out the metadata records rather than a string representation of the object name when calling the toString() method on a class object.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/models/JobMetadata.java
+++ b/tony-core/src/main/java/com/linkedin/tony/models/JobMetadata.java
@@ -35,20 +35,12 @@ public class JobMetadata {
   public static JobMetadata newInstance(YarnConfiguration conf, String histFileName) {
     String histFileNoExt = histFileName.substring(0, histFileName.indexOf('.'));
     String[] metadata = histFileNoExt.split("-");
-    Builder metadataBuilder = new Builder()
-        .setId(metadata[0])
-        .setStarted(Long.parseLong(metadata[1]))
-        .setConf(conf);
+    Builder metadataBuilder = new Builder().setId(metadata[0]).setStarted(Long.parseLong(metadata[1])).setConf(conf);
     if (histFileName.endsWith(Constants.INPROGRESS)) {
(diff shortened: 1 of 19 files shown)


tony-core/src/main/java/com/linkedin/tony/models/JobMetadata.java
```diff
@@ -142,4 +134,11 @@ public void setId(String id) {
   public void setUser(String user) {
     this.user = user;
   }
-}
+
+  @Override
+  public String toString() {
+    return "\nMetadata{" + "id: " + id + ",\n joblink: " + jobLink + ",\n configlink: " + configLink + ",\n rmLink: "
```
erwa: I would suggest removing all the `\n`s in the returned string. It's more standard in my experience to keep the default toString() representation a single line and let the user add newlines if they so desire (example: if iterating over a list of JobMetadata, print one JobMetadata per line). Also, when you use `grep`, by default it returns matching lines, so it's generally more convenient when an object's entire string representation is on one line. If a user really wants to print each part on a separate line, they can always call getId(), getJobLink(), etc. separately and construct their own custom string.

You can also merge the first two string segments together: `"Metadata{" + "id: "` --> `"Metadata { "id: "`

tony-portal/app/controllers/JobsMetadataPageController.java
```diff
@@ -16,7 +18,15 @@ public JobsMetadataPageController(CacheWrapper cacheWrapper) {
...
   }
 
+
   public Result index() {
-    return ok(views.html.metadata.render(cache.asMap().values()));
+
+    return ok(views.html.metadata.render(cache.asMap().values()
+        .stream()
+        .sorted(Comparator.comparingLong(JobMetadata::getCompleted).reversed())
+        .collect(Collectors.toList())));
```
erwa: Nice use of streams!

PR #383, opened 2019-10-02
circleci

oliverhu:
Now it is running: https://circleci.com/gh/linkedin/TonY/6


erwa:
How come you decided to switch from TravisCI to CircleCI? Should we remove the .travis.yml file?

oliverhu:
yah, we should. CircleCI is a more stable and you could ssh into the container that runs the job to debug there @erwa

PR #389, opened 2019-10-03
Prepare for a new release

oliverhu:
bump version for new maven release


oliverhu:
- Scalability fixes to support 200+ concurrent executors.
- UX fix to print out am log url and logging
- Support MXNet
- Fix port race condition problem

PR #391, opened 2019-10-08
Concurrent containerIdMap updates will corrupt AM state, causing it t…

hungj:
…o hang

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -1044,7 +1044,7 @@ public class ApplicationMaster {
 
     @Override
     public void onNodesUpdated(List<NodeReport> list) {
-      LOG.info("onNodesUpdated called in RMCAllbackHandler");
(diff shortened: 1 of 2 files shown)


erwa:
Just curious, can you explain why using a regular HashMap might cause corruption?

Side note: I think you could still keep the interface `Map<>` rather than using the specific `ConcurrentHashMap` as the external type of `containerIdMap`.

hungj:
Hey @erwa , issue is that in https://github.com/linkedin/TonY/pull/387, we changed ContainerLauncher to run in a thread pool so that containers are launched concurrently. This means TonySession#addContainer is called concurrently so stuff is added to containerIdMap concurrently. So we observed two containers being added to this map at the same time, which regular hashmap cannot handle (one of the containerId keys got overwritten, even though they were different containerIds), hence the change to ConcurrentHashMap.

For the interface, I chose to use ConcurrentHashMap for the declaration since that's the contract we want containerIdMap to adhere to (a generic Map will not suffice).

erwa:
Ah, thanks for the explanation. That makes sense. Your choice of using ConcurrentHashMap makes sense in that context as well.

PR #394, opened 2019-10-12
Support configuring node label per job type

hungj:


tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -944,7 +944,7 @@ private void setupContainerCredentials() throws IOException {
     Priority priority = Priority.newInstance(request.getPriority());
     Resource capability = Resource.newInstance((int) request.getMemory(), request.getVCores());
     Utils.setCapabilityGPU(capability, request.getGPU());
-    AMRMClient.ContainerRequest containerRequest = new AMRMClient.ContainerRequest(capability, null, null, priority);
+    AMRMClient.ContainerRequest containerRequest = new AMRMClient.ContainerRequest(capability, null, null, priority, true, request.getNodeLabelsExpression());
```
oliverhu: is node-label null handled here?

hungj: Yeah, if node label is unspecified in ContainerRequest (i.e. what it was before this pr) it will route to default partition/app node label

PR #403, opened 2019-11-07
TonY should be able to launch jobs using dependency graph

UWFrankGu:


erwa:
Very interesting feature! Is there an issue or doc explaining more the motivation and use case for this feature? Is this to support things like the distributed hash table in slide 28 here?: https://www.slideshare.net/ssuser72f42a/scaling-deep-learning-on-hadoop-at-linkedin

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
```diff
@@ -185,6 +188,15 @@ public int getNumFailedTasks() {
     return (int) jobTasks.values().stream().flatMap(arr -> Arrays.stream(arr)).filter(task -> task.isFailed()).count();
   }
 
+  /** Number of expected tasks that have been scheduled at current time **/
+  public int getNumExpectedTasks() {
+    return numExpectedTasks;
+  }
+
+  public void addExpectedTask() {
+    numExpectedTasks++;
```
oliverhu: any potential race condition on this ?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+      for (TensorFlowContainerRequest containerRequest : containersRequests) {
+        if (!isSubGraphDAG(containerRequest, containersRequestsMap)) {
+          return false;
+        }
+      }
+
+      return true;
+    }
+
+    private boolean isSubGraphDAG(TensorFlowContainerRequest root, final Map<String, TensorFlowContainerRequest> lookupMap) {
```
oliverhu: nit: `isSubgraphDAG`

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
     return rpcServer;
   }
 
+  private class TaskScheduler {
```
oliverhu: Move this to a new file? AM class is large enough, it would also be easier to unit test TaskScheduler if it is in another class (w/ proper constructors for testing)

oliverhu: we need some unit testing for this  class as well..

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+      while (waitingRequestItr.hasNext()) {
+        TensorFlowContainerRequest waitingRequest = waitingRequestItr.next();
+        if (checkDependencySatisfied((waitingRequest))) {
+          waitingRequestItr.remove();
+          scheduleJob(waitingRequest);
+        }
+      }
+    }
+
+    private boolean isDAG(final List<TensorFlowContainerRequest> containersRequests) {
```
oliverhu: interesting.. are you using DFS to check if it is a DAG ? We need some unit tests for this,

tony-core/src/main/java/com/linkedin/tony/tensorflow/TensorFlowContainerRequest.java
```diff
@@ -13,16 +16,18 @@
   private int gpu;
   private String jobName;
   private String nodeLabelsExpression;
+  private List<String> dependsOn;
 
   public TensorFlowContainerRequest(String jobName, int numInstances, long memory, int vCores, int gpu, int priority,
```
oliverhu: we should rename this to just JobContainerRequest, can be next PR

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
ashahab: definitely. Also add documentation on why we have capabilities here that are duplicating Azkaban's DAG management capabilities

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
     return rpcServer;
   }
 
+  private class TaskScheduler {
+    private Map<TensorFlowContainerRequest, Map<String, Integer>> taskDependencyMap = new HashMap<>();
```
hungj: Perhaps add a comment describing what this map contains?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+
+      while (!stack.isEmpty()) {
+        TensorFlowContainerRequest node = stack.pop();
+        if (visited.contains(node)) {
+          return false;
+        }
+
+        visited.add(node);
+
+        List<TensorFlowContainerRequest> neighbours = lookupMap.entrySet().stream()
```
hungj: neighbours -> dependencies?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+            .filter(x -> node.getDependsOn().contains(x.getKey()))
+            .map(Map.Entry::getValue)
+            .collect(Collectors.toList());
+
+        for (TensorFlowContainerRequest neighbour : neighbours) {
```
hungj: neighbour -> dependency

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+          scheduleJob(waitingRequest);
+        }
+      }
+    }
+
+    private boolean isDAG(final List<TensorFlowContainerRequest> containersRequests) {
+      Map<String, TensorFlowContainerRequest> containersRequestsMap = new HashMap<>();
+      containersRequests.forEach(x -> containersRequestsMap.put(x.getJobName(), x));
+
+      for (TensorFlowContainerRequest containerRequest : containersRequests) {
```
hungj: Seems this might be optimizable by creating the `visited` set here, passing it through to isSubgraphDAG and checking in this for loop if we've seen it before

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+    private Map<TensorFlowContainerRequest, Map<String, Integer>> taskDependencyMap = new HashMap<>();
+    boolean dependencyCheckPassed = true;
+
+    private void scheduleTasks() {
+      session.setResources(yarnConf, hdfsConf, localResources, containerEnv, hdfsClasspath);
+      final List<TensorFlowContainerRequest> requests = session.getContainersRequests();
+
+      if (!isDAG(requests)) {
+        LOG.error("TonY execution graph does not form a DAG, exiting.");
+        session.setFinalStatus(FinalApplicationStatus.FAILED, "TonY task scheduling failed..");
```
hungj: App failed due to it not being a DAG, let's add this in the error message here

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+          scheduleJob(request);
+        }
+      }
+    }
+
+    private void buildTaskDependency(List<TensorFlowContainerRequest> requests) {
+      for (TensorFlowContainerRequest request : requests) {
+        for (String dependsOn : request.getDependsOn()) {
+          if (!dependsOn.isEmpty()) {
+            taskDependencyMap.putIfAbsent(request, new HashMap<>());
```
hungj: We only construct taskDependencyMap here, so I think you can take this bit out, and in the next line: `Map<String, Integer> dependenciesForTask = new HashMap<>();`

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+      final List<TensorFlowContainerRequest> requests = session.getContainersRequests();
+
+      if (!isDAG(requests)) {
+        LOG.error("TonY execution graph does not form a DAG, exiting.");
+        session.setFinalStatus(FinalApplicationStatus.FAILED, "TonY task scheduling failed..");
+        dependencyCheckPassed = false;
+        return;
+      }
+
+      // build task dependency graph
```
hungj: Delete this comment and rename the method `buildTaskDependencyGraph`?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -780,6 +759,147 @@ private ApplicationRpcServer setupRPCService(String hostname) {
...
+      AMRMClient.ContainerRequest containerAsk = setupContainerRequestForRM(request);
+      String jobName = request.getJobName();
+      if (!jobTypeToContainerRequestsMap.containsKey(jobName)) {
+        jobTypeToContainerRequestsMap.put(jobName, new ArrayList<>());
+        jobTypeToContainerResources.put(jobName, getContainerResources(jobName));
+      }
+      jobTypeToContainerRequestsMap.get(request.getJobName()).add(containerAsk);
+      for (int i = 0; i < request.getNumInstances(); i++) {
+        amRMClient.addContainerRequest(containerAsk);
+        session.addExpectedTask();
```
hungj: can we just call the method `setNumExpectedTasks` and do `setNumExpectedTasks(request.getNumInstances())` once outside the forloop?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
UWFrankGu: A task may depend on multiple other tasks, therefore some dependency might already exist in the taskDependencyMap.

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
oliverhu: we are not duplicating Azkaban, let's not mention that in documentation... have some mini DAG engine in any framework is expected.

erwa:
This PR is to support the Needle use case, right?

UWFrankGu:
> This PR is to support the Needle use case, right?

Right

erwa:
Good stuff.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -366,6 +374,15 @@ public static void addEnvironmentForResource(LocalResource resource, FileSystem
               TonyConfigurationKeys.DEFAULT_GPUS);
       String nodeLabel = conf.get(TonyConfigurationKeys.getNodeLabelKey(jobName));
 
+      // Disabled building of DAG graph, for now we only support simple two stage tasks
```
hungj: Let's delete line 377 and 378

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -351,10 +352,17 @@ public static void addEnvironmentForResource(LocalResource resource, FileSystem
...
     Set<String> jobNames = getAllJobTypes(conf);
-    Map<String, TensorFlowContainerRequest> containerRequests = new HashMap<>();
+    Set<String> untrackedJobTypes = Arrays.stream(getUntrackedJobTypes(conf)).collect(Collectors.toSet());
+    Map<String, JobContainerRequest> containerRequests = new HashMap<>();
     int priority = 0;
+
+    List<String> prepareStageTasks = new ArrayList<>(conf.getTrimmedStringCollection(TonyConfigurationKeys.APPLICATION_PREPARE_STAGE));
+    List<String> trainingStageTasks = new ArrayList<>(conf.getTrimmedStringCollection(TonyConfigurationKeys.APPLICATION_TRAINING_STAGE));
+    ensureStagedTasksIntegrity(prepareStageTasks, trainingStageTasks, jobNames);
+    List<String> tasksToDependOn = prepareStageTasks.stream().filter(x -> !untrackedJobTypes.contains(x)).collect(Collectors.toList());
```
hungj: I guess the intent here is we don't want to depend on tasks which will never terminate. Is this the api we want to expose though? Ithink it's a bit confusing here, the user would need to specify "dependent" jobs (both terminating and non-terminating), then specify the non-terminating one in non-tracked job types.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -374,13 +391,34 @@ public static void addEnvironmentForResource(LocalResource resource, FileSystem
...
+      prepareStageTasks.addAll(CollectionUtils.subtract(allJobTypes, trainingStageTasks));
+      LOG.warn("Found no prepare stage tasks, auto-filling with: " + Arrays.toString(prepareStageTasks.toArray()));
+    } else if ((!prepareStageTasks.isEmpty() && trainingStageTasks.isEmpty())) {
+      trainingStageTasks.addAll(CollectionUtils.subtract(allJobTypes, prepareStageTasks));
+      LOG.warn("Found no training stage tasks, auto-filling with: " + Arrays.toString(trainingStageTasks.toArray()));
+    } else if (prepareStageTasks.isEmpty() && trainingStageTasks.isEmpty()) {
+      return;
+    }
+
+    if (prepareStageTasks.size() + trainingStageTasks.size() != allJobTypes.size()) {
```
hungj: See the other comment, Do we need to enforce this invariant? IMO we should just enforce the user's configured 'training' tasks depend on user's configured 'prepare' tasks, everything else can just be scheduled as normal

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
UWFrankGu: I think the api we expose is prepare-stage and training-stage, there's no notion of what task depend on what, but rather what tasks are launched in what stage. Once we open up the dependsOn api, these code can be removed.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
UWFrankGu: I think this makes sense, I'll remove the training-stage api. Just having prepare-stage is enough

hungj: Discussed offline...it's best to enforce this ( prepare jobs + training jobs = all jobs). Otherwise if a job is  misconfigured (e.g. a job that was supposed to be in training phase, but was not configured as such) will be allocated first and will sit idle (wasting cluster resources).

Issue #405, opened 2019-12-11
Handle untracked job type crash by failing the application.


goyalankit:
There have been cases where if the 'ps' crashes, then the worker
    processes hang forever waiting for establishment of session with the
    ps.

goyalankit:
#406

PR #406, opened 2019-12-11
Handle untracked job type crash by failing the application for #405

goyalankit:
There have been cases where if the 'ps' crashes, then the worker
processes hang forever waiting for establishment of session with the
ps.

This change kills the application if the ps or any of the untracked
processes exits in a FAILED state. This excludes cases where ps
exits with an exit code of 0 or was killed explicitly by the AM.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -156,6 +156,9 @@ public class ApplicationMaster {
   /** Preprocessing job **/
   private boolean enablePreprocessing = false;
 
+  /** Untracked jobs **/
+  private boolean untrackedTaskFailed = false;
+
   /** Lifecycle control **/
   private long appTimeout;
 
@@ -596,6 +599,11 @@ public class ApplicationMaster {
         break;
       }
 
+      if (this.untrackedTaskFailed) {
+        LOG.error("One of the untracked tasks has failed with a non-zero exit code.");
+        break;
+      }
+
       if (!this.scheduler.dependencyCheckPassed) {
         LOG.info("Terminating application due to failure to load dependency graph");
         break;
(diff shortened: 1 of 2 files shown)


tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -596,6 +599,11 @@ private boolean monitor() {
         break;
       }
 
+      if (this.untrackedTaskFailed) {
```
oliverhu: I don't see you update final state based on this?

goyalankit: That's being done in the [onTaskCompleted](https://github.com/linkedin/TonY/blob/master/tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java#L263) callback in the TonySession. 
In addition, it's also being updated again within the [monitor method](https://github.com/linkedin/TonY/blob/master/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java#L626
), as soon as we exit the loop:

tony-core/src/test/java/com/linkedin/tony/TestTonyE2E.java
```diff
@@ -414,6 +414,38 @@ public void testTonyClientCallbackHandler() throws IOException, ParseException {
...
+        "--conf", "tony.application.untracked.jobtypes=ps"
+      });
+    client.addListener(handler);
+    int exitCode = client.start();
+    Assert.assertEquals(exitCode, -1);
+    client.removeListener(handler);
+    Assert.assertEquals(handler.getTaskInfoSet().size(), 2);
+    for (TaskInfo taskInfo : handler.getTaskInfoSet()) {
+      String name = taskInfo.getName();
+      // Should
```
hungj: what's this comment?

goyalankit: woops.. fixed it.

goyalankit:
fixed #405

PR #409, opened 2019-12-13
Add support for shiv/pex python packages

goyalankit:
Zipping Venv is not a very reliable way of packaging python as it ships interpreter in the zip file which may or may not work with .so available on the target host. So add support for running executables from other package managers like shiv or pex.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -204,6 +204,15 @@ public class TonyClient implements AutoCloseable {
           new Path(pythonVenv), Constants.PYTHON_VENV_ZIP, tonyConf, fs, LocalResourceType.FILE, TonyConfigurationKeys.getContainerResourcesKey());
     }
 
+    // If the user has provided python binary path without the python virtual env
+    // and the binary path exists, then upload it.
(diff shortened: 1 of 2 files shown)


tony-azkaban/src/main/java/com/linkedin/tony/azkaban/TonyJob.java
```diff
@@ -157,6 +157,11 @@ protected String getMainArguments() {
       args.append(" " + TonyJobArg.PYTHON_VENV.tonyParamName + " " + pythonVenv);
     }
 
+    String interpreterPath = getJobProps().getString(TonyJobArg.INTERPRETER_PATH.azPropName, null);
```
oliverhu: just call executor path ?

goyalankit:
As discussed offline, it's better to not introduce a new param and re-purpose the python binary path. So updating the pull request with the new change.

goyalankit:
@oliverhu can you please take a look at the diff again? thanks

oliverhu:
@goyalankit we don't this PR any more from our discussion?

goyalankit:
yep. Closing this.
---

We can still support this use case by using tony.containers.resources to upload the shiv/pex executable and then using the python binary path or the executes command to use it.

Issue #420, opened 2020-01-16
Add option to return SUCCEED when training is completed with some failed job tasks


charliechen211:
![image](https://user-images.githubusercontent.com/5773445/72504267-74800780-3878-11ea-8293-2e0cbc78f5bf.png)

App will be stopped when chief worker (or some certain jobtype), and continue training when other job task fails. It looks reasonable.

![image](https://user-images.githubusercontent.com/5773445/72504503-e5272400-3878-11ea-886b-5a7c1d7d77f6.png)

But in this circumstance, training is finished and return FAILED on YARN. 

I think it's a contradiction. 
If we will continue training after one worker fails, and training is finally completed, why should it return FAILED? 
Users would feel confused, and don't know what they should do next. Rerun training for more accurate model, or use this model with a FAILED status?

They could config #412 if they needs fail fast, while return SUCCEED when training is completed if they needs training to continue.

I believe it is better to add an extra option for whether return SUCCEED or FAILED when training is completed with some failed job tasks, which user can specify on a per-job basis.

morenn520:
@oliverhu

oliverhu:
Agreed, but I am refrained from making the APIs over complicated, any thought on how the configurations would look like?

morenn520:
@oliverhu 
emmm...I made a PR, in which the configuration naming is not very good.
https://github.com/linkedin/TonY/pull/421/files

I also agreed that it's better to simplify the APIs, or we can cancel this configuration and all completed application with failure tasks will return SUCCEEDED, which is more reasonable.

morenn520:
CI failed with following logs:
Too long with no output (exceeded 10m0s): context deadline exceeded

Link: https://circleci.com/gh/morenn520/TonY/2?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Maybe test has been run too long?

oliverhu:
Not sure, will take a look tmr

oliverhu:
Assign to Ankit to follow up on this issue

PR #421, opened 2020-01-17
 Add option to return SUCCEED when training is completed with some failed job tasks

morenn520:
Description: https://github.com/linkedin/TonY/pull/420

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
@@ -246,6 +246,10 @@ public class TonyConfigurationKeys {
   // Job types that we will short circuit when it failed
   public static final String STOP_ON_FAILURE_JOBTYPES = TONY_APPLICATION_PREFIX + "stop.on.failure.jobtypes";
 
+  // Whether to return FAILED when training is completed with some failed job tasks
+  public static final String FAILED_ON_COMPLETED_WITH_FAILURE_ENABLED = TONY_APPLICATION_PREFIX + "failed_on_complete_with_failure.enabled";
+  public static final boolean DEFAULT_FAILED_ON_COMPLETED_WITH_FAILURE_ENABLED = true;
+
   // Training chief configuration
   public static final String CHIEF_PREFIX = TONY_PREFIX + "chief.";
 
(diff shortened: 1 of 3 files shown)


oliverhu:
Triggered another run of the job.

oliverhu:
CI complaints I don’t have permission to rebuild it. @morenn520 can you push an empty commit to trigger another build ?

tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
```diff
@@ -246,6 +246,10 @@ public static String getContainerDockerKey() {
   // Job types that we will short circuit when it failed
   public static final String STOP_ON_FAILURE_JOBTYPES = TONY_APPLICATION_PREFIX + "stop.on.failure.jobtypes";
 
+  // Whether to return FAILED when training is completed with some failed job tasks
+  public static final String FAILED_ON_COMPLETED_WITH_FAILURE_ENABLED = TONY_APPLICATION_PREFIX + "failed_on_complete_with_failure.enabled";
```
oliverhu: Merge with the logic to handle STOP_ON_FAILURE_JOBTYPES? 
Also use a-b-c instead of a_b_c
And add a test

morenn520: They could config STOP_ON_FAILURE_JOBTYPES if they needs fail fast, while return SUCCEED when training is completed if they needs training to continue by FAILED_ON_COMPLETED_WITH_FAILURE_ENABLED.

So I don't think FAILED_ON_COMPLETED_WITH_FAILURE_ENABLED could be merged with STOP_ON_FAILURE_JOBTYPES.

morenn520: For test, I should run a job where one worker will be failed and others continue to complete the training finally. And I don't find some test case like this in TonY. @oliverhu Any thought or suggestion on UT?

oliverhu: on a second thought,  we could refactor and follow the convention of `tony.x.stop_on_failure` `tony.x.fail_app_on_failure`, basically use this to form the table we discussed in the issue, thoughts?
Check the E2E tests, they mimic real life tony jobs.

oliverhu:
moved to discuss in #421

PR #426, opened 2020-02-10
Support to specify application type

yuanzac:
We add a new parameter named tony.application.type which to specify application type for jobs.


yuanzac:
@oliverhu,
Thanks for the review. Failed tests are fixed.

PR #427, opened 2020-02-11
Fixing AM RPC port binding issue

arde171:
Fixing AM RPC port allocation to avoid port binding issue.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -385,10 +385,7 @@ public class ApplicationMaster {
 
     // Setup application RPC server
     String amHostname = Utils.getCurrentHostName();
-    ServerSocket amRpcSocket = new ServerSocket(0);
-    int amRpcPort = amRpcSocket.getLocalPort();
-    amRpcSocket.close();
-    applicationRpcServer = setupRPCService(amHostname, amRpcPort);
+    applicationRpcServer = setupRPCService(amHostname);
     containerEnv.put(Constants.AM_HOST, amHostname);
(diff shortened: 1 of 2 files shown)


oliverhu:
how would this solve the problem ?

goyalankit:
> how would this solve the problem ?

AM was using a random number to allocate port. We can use serversocket with port 0 to get OS to allocate a port.

oliverhu:
@goyalankit  I see.. that was dumb, I thought we used port 0 to get OS

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -385,7 +385,10 @@ private boolean prepare() throws IOException {
 
     // Setup application RPC server
     String amHostname = Utils.getCurrentHostName();
-    applicationRpcServer = setupRPCService(amHostname);
+    ServerSocket amRpcSocket = new ServerSocket(0);
+    int amRpcPort = amRpcSocket.getLocalPort();
+    amRpcSocket.close();
```
oliverhu: move this setupRPCService() ?

arde171: @oliverhu  ServerSocket may throw IOException https://docs.oracle.com/javase/8/docs/api/java/net/ServerSocket.html
So, we have to handle it cautiously. If we move this code to setupRPCService() then 
we need to handle throw exception there. Is it OK?

arde171: @oliverhu  As discussed, I moved the port allocation code inside the ApplicationRpcServer constructor.

PR #428, opened 2020-02-14
Added fix to return successful training with some failed worker jobs 

arde171:
Added a configuration parameter called “tony.application.fail.on.worker.failure.enabled” with default value “false” to continue training if some worker jobs failed.  If it configured  “true” then it would immediately stop training if a worker failed.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
@@ -251,6 +251,10 @@ public class TonyConfigurationKeys {
   // Job types that we will short circuit when it failed
   public static final String STOP_ON_FAILURE_JOBTYPES = TONY_APPLICATION_PREFIX + "stop.on.failure.jobtypes";
 
+  // Tony configuration to return failure when a worker failed
+  public static final String FAIL_ON_WORKER_FAILURE_ENABLED = TONY_APPLICATION_PREFIX + "fail.on.worker.failure.enabled";
+  public static final boolean DEFAULT_FAIL_ON_WORKER_FAILURE_ENABLED = false;
+
   // Training chief configuration
   public static final String CHIEF_PREFIX = TONY_PREFIX + "chief.";
 
--- a/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
@@ -249,10 +249,10 @@ public class TonyConfigurationKeys {
   public static final String UNTRACKED_JOBTYPES_DEFAULT = "ps";
 
   // Job types that we will short circuit when it failed
-  public static final String STOP_ON_FAILURE_JOBTYPES = TONY_APPLICATION_PREFIX + "stop.on.failure.jobtypes";
+  public static final String STOP_ON_FAILURE_JOBTYPES = TONY_APPLICATION_PREFIX + "stop-on-failure-jobtypes";
 
   // Tony configuration to return failure when a worker failed
-  public static final String FAIL_ON_WORKER_FAILURE_ENABLED = TONY_APPLICATION_PREFIX + "fail.on.worker.failure.enabled";
+  public static final String FAIL_ON_WORKER_FAILURE_ENABLED = TONY_APPLICATION_PREFIX + "fail-on-worker-failure-enabled";
   public static final boolean DEFAULT_FAIL_ON_WORKER_FAILURE_ENABLED = false;
 
   // Training chief configuration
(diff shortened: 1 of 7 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
```diff
@@ -251,6 +251,10 @@ public static String getContainerDockerKey() {
   // Job types that we will short circuit when it failed
   public static final String STOP_ON_FAILURE_JOBTYPES = TONY_APPLICATION_PREFIX + "stop.on.failure.jobtypes";
 
+  // Tony configuration to return failure when a worker failed
+  public static final String FAIL_ON_WORKER_FAILURE_ENABLED = TONY_APPLICATION_PREFIX + "fail.on.worker.failure.enabled";
```
UWFrankGu: I think we want to use the '.' delimiter for creating a hierarchy. In this case it would be better to follow the convention: "fail-on-worker-failure-enabled".

tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
```diff
@@ -251,6 +251,10 @@ public static String getContainerDockerKey() {
   // Job types that we will short circuit when it failed
   public static final String STOP_ON_FAILURE_JOBTYPES = TONY_APPLICATION_PREFIX + "stop.on.failure.jobtypes";
```
UWFrankGu: ditto

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
```diff
@@ -258,11 +259,14 @@ public void onTaskCompleted(String jobName, String jobIndex, int exitCode) {
...
     // In addition, we support custom jobtype config`tony.application.stop.on.failure.jobtypes`,
     // and we short circuit when that jobtype instance failed.
+    // We added a configuration parameter called `tony.application.fail.on.worker.failure.enabled` with default
+    // value “false” to continue training if some worker jobs failed. If we configured it “true” then it would
+    // immediately stop training if a worker failed.
     if (exitCode != ContainerExitStatus.SUCCESS && exitCode != ContainerExitStatus.KILLED_BY_APPMASTER) {
-      if (isChief(jobName, jobIndex) || shouldStopOnFailure(jobName)) {
+      if (isChief(jobName, jobIndex) || shouldStopOnFailure(jobName) || isFailOnWorkerFailure()) {
         trainingFinished = true;
+        setFinalStatus(FinalApplicationStatus.FAILED, "Exit status: " + exitCode);
```
UWFrankGu: Isn't this `FinalApplicationStatus` a duplicate from the one that `updatesessionStatus()` will give? We should handle `FinalApplicationStatus` only at the session-level but not at the task-level.

tony-core/src/main/java/com/linkedin/tony/TonyConfigurationKeys.java
arde171: ok

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
arde171: This was for short circuit the training in case of failure. In updatesessionStatus(),  it first checked the status  and return without scanning each task status.

UWFrankGu: I see, could you add a comment here on the short-circuiting

UWFrankGu: In this case, is it necessary to check `isFailOnWorkerFailure()` again in the `updateSessionStatus()`?

arde171: Yes, Incase it didn't catch it onTaskCompleted().

UWFrankGu: Could you add some comments on under what case it won't catch it here. Thanks

arde171: Done

tony-core/src/main/java/com/linkedin/tony/tensorflow/TonySession.java
```diff
@@ -306,8 +310,18 @@ public void updateSessionStatus() {
     }
 
     if (failureCount > 0) {
-      setFinalStatus(FinalApplicationStatus.FAILED,
-          "At least one job task exited with non-zero status, failedCnt=" + failureCount);
+      // if `tony.application.fail-on-worker-failure-enabled` is true and there is at least one failed worker,
+      // or if all worker jobs failed, we finally return "FAILED" otherwise, we return "SUCCEEDED" with some
+      // failure workers
```
UWFrankGu: I would also mention that the reason why we try to catch the failure again here is because when the AM killed the worker task, the `onTaskCompleted()` call won't be called to trigger the short-circuit

zuston:
The conf param of `fail-on-worker-failure-enabled` makes me confused, because of confusion with tensorflow's **worker** role.

Maybe `fail-on-task-failure-enabled` will be better.

Issue #441, opened 2020-05-29
tony-portal distribution is not published correctly


haibchen:
tony-portal is a play application that is distributed as a zip file. But right now, jars and javadocs are released to maven instead.

PR #442, opened 2020-05-29
fix publication of tony-portal distribution

haibchen:
To address issue #441 

Tested with 'gradle publishToMavenLocal' and 'gradle publish' to Nexus Staging repo. Both worked fine.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -159,6 +159,7 @@ subprojects {
   publishing {
     publications {
       mavenJava(MavenPublication) {
+        artifactId = "${project.name}"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -159,10 +159,13 @@ subprojects {
   publishing {
     publications {
       mavenJava(MavenPublication) {
-        from components.java
-        artifactId = "${project.name}"
```
goyalankit, line 163: Is this not required for either?

haibchen: added back

PR #449, opened 2020-06-10
Sets an environment variable when restarting due to preemption

StanfordMCP:


tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -353,6 +353,10 @@ private boolean run(String[] args) throws IOException {
       // Prepare for retryCount.
       reset();
       LOG.info("Retrying, remaining retry count" + amRetryCount);
+
+      //Set an environment variable when restarting due to preemption
+      containerEnv.put(Constants.AM_RETRY_COUNT, Integer.toString(amRetryCount));
```
oliverhu, line 358: This is the remaining retry count, we need the retry id (total-remaining) here.

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -353,7 +354,13 @@ private boolean run(String[] args) throws IOException {
       // Prepare for retryCount.
       reset();
       LOG.info("Retrying, remaining retry count" + amRetryCount);
+
       amRetryCount -= 1;
+      numAMRetries += 1;
```
oliverhu, line 359: prefer to start from 0, move the +1 after next line

StanfordMCP: Ok, so set it to 0 within the scope of the method? Not line 118?

oliverhu: 118 should be ok, just move this to the nextline

oliverhu:
👍  @StanfordMCP
 please also update the internal ticket with the details

StanfordMCP:
Updated the internal ticket and closed it.

PR #366, opened 2019-09-04
TaskExecutor should release port immediately prior to launching shell.

burgerkingeater:
Problem description: https://github.com/linkedin/TonY/issues/365.

Unit tests to be added as follow up.

Issue #365, opened 2019-09-04
TaskExecutor should release port immediately prior to launching shell


hungj:
If there's a large TonY app (e.g. >200 executors), it's possible one TaskExecutor sets up ports long before another. By this time the first TaskExecutor will have acquired and released a port, and the longer the setup time between two executors, the more likely some other TaskExecutor (by a different application) will try to setup on the same port.

The fix is for TaskExecutor to hold onto the port up until immediately before the underlying TF process is launched (or at least hold on to the port until the TaskExecutor receives the cluster spec, meaning all TaskExecutors in the same job have been set up) so that we block this port while the rest of the TaskExecutors are being allocated/setting up.

burgerkingeater:
@hungj can we close this issue given it's resolved?

erwa:
Fixing this issue is a good way to mitigate the chance of a race condition causing port contention. However, the race condition still exists, though the window where it might happen has been reduced. I think we could eliminate the race condition entirely between different TaskExecutors running on the same machine as follows:

* rather than finding an arbitrary free port, each TaskExecutor could iterate over a range of ports (either a randomly chosen range or something specified in config)
* write a temp file `port_XXX` to some common location (could be specified in tony-site.xml)
* if write succeeds, try to open socket on port (as sanity check). Else, go back to the previous step, trying another port
* release port, set the port environment variable
* spawn the process
* when TaskExecutor exits, temp file `port_XXX` should be cleaned up automatically

I think this would work, assuming:
* other processes and applications are not using the same ports (shouldn't be a problem in general, since MR and Spark have well-defined ports or are using dynamic ports (the range of dynamic ports should be configurable, and as long as the TaskExecutors use a port range which doesn't overlap with the dynamic port range, there shouldn't be a conflict))
* the TaskExecutor process isn't SIGKILL'd (if it is, temp file clean up won't happen). I forget if YARN tries SIGTERM first before sending SIGKILL when a user kills an application or when nmClientAsync.stopContainerAsync is called. For the latter, we could replace the stopContainerAsync() call with calling our own TaskExecutor RPC endpoint telling the TaskExecutor to shutdown and do any necessary clean-up. For the former, temp `port_XXX` files left behind just means those ports won't be used by TaskExecutors until the file is cleaned up (but we could ensure they are cleaned up by a tmpwatch script).

oliverhu:
+1 to @erwa , this sounds like a good enough synchronization mechanism for now.

PR #453, opened 2020-07-07
Hold reserved ports till task finishes to avoid possible port conflict in TaskExecutor

burgerkingeater:
Even with fix in https://github.com/linkedin/TonY/pull/366, task executor is still subject to **port conflict** where port could be grabbed by other process on the host before task process launches and after ports are closed. To address such conflict, following mechanism is applied:
Task executor will iterate over defined port range and check a port's availability based on 1. socket API **and** 2. the existence of corresponding port file. If socket can be created with the port and corresponding port file doesn't exist, a temp file tracking that port will be created under a shared directory(specified by _java.io.tmpdir_) and that port will be reserved for the process. The file will be deleted upon the process finishes.

A similar approach is also mentioned in https://github.com/linkedin/TonY/issues/365#issuecomment-529116493.

Note as mentioned in https://github.com/linkedin/TonY/pull/453#issuecomment-655914054, if other non-tony processes grab the port during the gap after ports are closed and before task process launches, it's still subject to the risk of port conflict.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
@@ -70,6 +70,8 @@ public class TaskExecutor {
   private final ScheduledExecutorService scheduledThreadPool = Executors.newScheduledThreadPool(2);
   private int numFailedHBAttempts = 0;
   private MLFramework framework;
+  // Port numbers range from 0 to 65535, but 0 to 1023 are reserved for privileged
+  // services and designated as well-known ports.
   public static final Range<Integer> PORT_RANGE = Range.between(1024, 65535);
   public static final String PORT_FILE_PREFIX = "___port___";
   public static final java.nio.file.Path PORT_FILE_DIR = Paths.get(System.getProperty("java.io"
--- a/tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
@@ -12,7 +12,6 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
-import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.HashMap;
@@ -102,7 +101,7 @@ public class TaskExecutor {
       // in case task executor is killed abruptly, delete port file
       file.deleteOnExit();
       boolean fileCreated = file.createNewFile();
-      if(!fileCreated) {
+      if (!fileCreated) {
         LOG.debug("Port " + port + " is not available");
         serverSocket.close();
         return null;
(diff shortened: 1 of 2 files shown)


ChengbingLiu:
How often do you encounter a port issue? If the ServerSocket(0) finds a random available port, then it is very unlikely to have a port conflict. If it happens too often, maybe the selected port is not (fully) randomized?

burgerkingeater:
@ChengbingLiu thanks for your comment! Not sure about the how port allocation algorithm works in ServerSocket, but the tests shown it's not fully randomized and the distribution is pretty dense - it's always picking up a port within a certain range(>= 60000). We have seen a non-significant gap (30 secs) between port release and process starts, so there's some chance port conflict could happen.

ChengbingLiu:
@burgerkingeater I also see the gap between port release and process starts (port bind inside the process, to be precise), and it depends on the specific program (could be longer or shorter).

I think the solution in this PR only works for a cluster with only TonY applications. Other kinds of application, or even system processes with bind behaviour can also cause a port issue. IMO, if YARN does not treat ports as allocatable resources, there could be hardly a solution that completely resolves this issue. Our main goal would be to reduce the probability of having a port issue as much as possible (by fully randomization, etc.).

A bit of code review: for most Linux system, port below 1024 are well-known ports, which only root user can listen on. So it's better to use 1024 as default range start.

burgerkingeater:
@ChengbingLiu you are right, it cannot work with other non-tony applications. I will note it in PR description as well.

- port below 1024 are well-known ports, which only root user can listen on. So it's better to use 1024 as default range start.
Thanks, will update the PR.

asamadiya:
@burgerkingeater 
Port 0 allocates a free ephemeral port from the range 49152 to 65535 (IANA standard). Linux may have a bigger range. Implementation of picking an ephemeral port is based on the specific kernel being used. It generally scans through ephemeral port range with some state from the previous round to make this scan efficient.

Your change is expanding this range and then implementing the port scanning logic. Do you have sufficient confidence that the current range of 15K+ ephemeral ports (more in linux) is not enough? This would mean, you have seen bind to port 0 fail on multiple occasions. If this is just for not conflicting the port with another process, I don't think this would help. Shortening the gap between port release and TF process port binding will better.

tony-core/src/main/java/com/linkedin/tony/TaskExecutor.java
```diff
@@ -68,33 +70,100 @@
...
+      return null;
+    } catch (Exception e) {
+      LOG.error(e);
+      return null;
+    }
+
+    try {
+      File file = PORT_FILE_DIR.resolve(PORT_FILE_PREFIX + port).toFile();
+      // in case task executor is killed abruptly, delete port file
+      file.deleteOnExit();
```
mridulm, line 104: This should be done after file is created, not before.

burgerkingeater:
@asamadiya @ChengbingLiu  thanks for your feedback. As I'm doing more research, I found tensorflow already supported port reuse in tf 2.3: https://github.com/tensorflow/tensorflow/pull/38705. It means if we can open a socket in tony with port reuse option enabled(supported in java 9+), and enable port reuse in tensorflow, this should cover our case.

burgerkingeater:
closing this PR since a better alternative is implemented in https://github.com/linkedin/TonY/pull/456

PR #465, opened 2020-09-15
Binding to the reusable port without accepting any inbound connection

burgerkingeater:
When TonY sets up the port with PORT_REUSE, it was using Netty whose default behavior is to bind to that port and starts a channel listening to the port(https://github.com/linkedin/TonY/pull/456). This could undesirably cause other remote TensorFlow workers to connect to the port within Tony process. 

The correct way is to just bind to that port but not accept any connection to it. 
Since Netty doesn't not allow port binding only, and Java 8 which TonY is running with doesn't have native support for port reuse, we are using python to achieve this.


tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
```diff
@@ -135,54 +119,39 @@ static ReusablePort create() throws IOException, InterruptedException {
...
-      } else {
-        LOG.info("Port " + port + " is no longer available.");
-        throw new BindException("Fail to bind to the port " + port);
-      }
-    } catch (Exception e) {
-      LOG.info("Reusable port allocation failed.", e);
-      close(bossGroup, future);
-      throw e;
+
+    String bindSocket = String.format("/export/apps/python/3.7/bin/python3 %s -p %s -t %s",
```
ashahab, line 140: Can the port be taken before this socket bind occurs?

burgerkingeater: the chance is very slim. gap between is < 1 s

ashahab:
Even so, is it possible for you to detect if some other process is holding
the port? Are there good error messages to emit for that?

burgerkingeater:
@ashahab yes, see https://github.com/linkedin/TonY/pull/465/files#diff-abc492838de5cbc22d337cc165fdba2fR209

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
```diff
@@ -107,23 +137,44 @@ private static int getAvailablePort() throws IOException {
...
 
+  private static boolean waitTillPortReserved(int port) {
+    assert RESERVE_PORT_SCRIPT_PATH != null;
+    Path parentPath = RESERVE_PORT_SCRIPT_PATH.getParent();
+    assert parentPath != null;
+    Path fileToWait = Paths.get(parentPath.toString(), port + PORT_FILE_NAME_SUFFIX);
+
+    int checkCount = 0;
+    int maxCheckCount = 5;
+    while (!Files.exists(fileToWait) && (checkCount++) < maxCheckCount) {
```
ashahab, line 164: This is probably fine for now but won't inotify be better(given the port can be reserved during the sleep): https://docs.oracle.com/javase/tutorial/essential/io/notification.html

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
```diff
@@ -107,23 +137,44 @@ private static int getAvailablePort() throws IOException {
...
+      try {
+        Duration checkInterval = Duration.ofSeconds(2);
```
ashahab, line 166: Pull it to the top, outside the while loop, and make "2" a constant

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
```diff
@@ -135,54 +186,35 @@ static ReusablePort create() throws IOException, InterruptedException {
...
-        throw new BindException("Fail to bind to the port " + port);
+    String socketBindingProcess = String.format("python %s -p %s -d %s",
+        RESERVE_PORT_SCRIPT_PATH, port, Duration.ofHours(1).getSeconds());
+
+    ProcessBuilder taskProcessBuilder = new ProcessBuilder("bash", "-c", socketBindingProcess);
+    taskProcessBuilder.redirectError(ProcessBuilder.Redirect.INHERIT);
+    taskProcessBuilder.redirectOutput(ProcessBuilder.Redirect.INHERIT);
+    if (isPortAvailable(port)) {
+      LOG.info("Starting process " + socketBindingProcess);
+      Process taskProcess = taskProcessBuilder.start();
```
ashahab, line 206: Document the change? That you are launching the python process first, and then waiting for its port-file to be created.

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
```diff
@@ -44,54 +44,84 @@
...
+    try {
+      // copy reserve_reusable_port.py from resource dir to a tmp dir
+      Path tempDir = Files.createTempDirectory("reserve_reusable_port");
+      tempDir.toFile().deleteOnExit();
+      try (InputStream stream = classloader.getResourceAsStream(reservePortScript)) {
+        Files.copy(stream, Paths.get(tempDir.toAbsolutePath().toString(), reservePortScript));
+      }
+      return Paths.get(tempDir.toAbsolutePath().toString(), reservePortScript);
+    } catch (IOException ex) {
+      return null;
```
annyan09023, line 64: what does this mean? Is it expected to have exception here?

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
```diff
@@ -44,54 +44,84 @@
...
+  ReusablePort(Process socketProcess, int port) {
+    this.socketProcess = socketProcess;
+    this.port = port;
   }
 
-  private static void close(EventLoopGroup loopGroup, ChannelFuture future) {
-    if (future != null && future.channel().isOpen()) {
-      future.channel().close().awaitUninterruptibly();
+  private static void killSocketBindingProcess(Process process) {
+    Preconditions.checkNotNull(process);
```
annyan09023, line 75: just some nits, I think you are mixing usage of Preconditions.checkNotNull and Objects.requireNonNull. I prefer always stick to one

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
```diff
@@ -107,23 +137,44 @@ private static int getAvailablePort() throws IOException {
...
-        LOG.info("port binding attempt " + (i + 1) + " failed.");
+        LOG.info("Port binding attempt " + (i + 1) + " failed.");
       }
     }
 
     throw new BindException("Unable to bind port after " + portBindingRetry + " attempt(s).");
   }
 
+  private static boolean waitTillPortReserved(int port) {
+    assert RESERVE_PORT_SCRIPT_PATH != null;
```
annyan09023, line 157: if assert works depending on system setting. Under some setting, assert doesn't work. Is it expected?

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
burgerkingeater: i did look at inotify but given it only takes few secs to create the file, this is simple yet effective(no need to spend much time waiting).

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
burgerkingeater: when exception is thrown, just return an invalid path.

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
burgerkingeater: it's to bypass annoying NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE detected by findbugs  plugin in tony, see https://stackoverflow.com/questions/23746395/getting-rid-of-np-null-on-some-path-from-return-value

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
ashahab: Ok but probably worth looking at this again.

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
burgerkingeater: thanks, updated, but i don't think make 2 a constant is better since it's only used once.

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
burgerkingeater: thanks, added the comment.

tony-core/src/main/java/com/linkedin/tony/ReusablePort.java
burgerkingeater: thanks, updated.

ashahab (approved):
LGTM

burgerkingeater:
@ashahab thanks

PR #477, opened 2020-11-04
Added catch for incompatible avro version

arde171:
I added **NoSuchMethodError** catch for handling incompatible avro version.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -326,7 +326,11 @@ public class ApplicationMaster {
         .setStarted(started)
         .setUser(user);
     JobMetadata metadata = metadataBuilder.build();
-    eventHandler.setUpThread(jobDir, metadata);
+
+    if (!eventHandler.setUpThread(jobDir, metadata)) {
+      return false;
+    }
+
     eventHandler.start();
     boolean succeeded;
     do {
--- a/tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
+++ b/tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
@@ -40,18 +40,23 @@ public class EventHandler extends Thread {
     myFs = fs;
   }
 
-  public void setUpThread(Path intermDir, JobMetadata metadata) {
+  public boolean setUpThread(Path intermDir, JobMetadata metadata) {
     if (intermDir == null) {
-      return;
+      return true;
     }
     inProgressHistFile = new Path(intermDir, HistoryFileUtils.generateFileName(metadata));
     try {
       out = myFs.create(inProgressHistFile);
       dataFileWriter.create(Event.SCHEMA$, out);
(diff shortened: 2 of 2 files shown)


tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -326,7 +326,11 @@ private boolean run(String[] args) throws IOException {
         .setStarted(started)
         .setUser(user);
     JobMetadata metadata = metadataBuilder.build();
-    eventHandler.setUpThread(jobDir, metadata);
+
+    if (!eventHandler.setUpThread(jobDir, metadata)) {
+      return false;
```
oliverhu, line 331: shall we fail the job if this fails? i don't think this is fatal

hungj (approved):

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
```diff
@@ -40,18 +40,23 @@ public EventHandler(FileSystem fs, BlockingQueue<Event> q) {
...
     if (intermDir == null) {
-      return;
+      return true;
     }
     inProgressHistFile = new Path(intermDir, HistoryFileUtils.generateFileName(metadata));
     try {
       out = myFs.create(inProgressHistFile);
       dataFileWriter.create(Event.SCHEMA$, out);
       LOG.info("Writing events to " + inProgressHistFile);
+    } catch (NoSuchMethodError e) {
```
ashahab, line 52: what happens if an exception is raised that's neither of these 2(NoSuchMethodError or IOException)? As much as it is an antipattern, should we return false for all `Exception` throws?

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
arde171: It returned false, when some exception occurred. I think it would be fine to avoid some unpredictable behavior latter (e.g., hanging for several hours without any allocation of resources)

tony-core/src/main/java/com/linkedin/tony/events/EventHandler.java
arde171: I'm little conservative with catching exceptions (legacy). I would prefer to catch exceptions that we can properly deal with otherwise just let it bubble up.

ashahab:
Yeah, I'm of two minds here. Not catching means we have to watch out for
the next exception that hangs TonY. And we can't really say we have fixed
all hanging issues.

arde171:
we can't ensure that it would fix all the hanging issues for different exceptions. Also, we don't know yet that other exceptions would also hang TonY. In this implementation, we just ensure that if there is a specific exception (e.g., NoSuchMethodError) occurred, it can handle gracefully.

oliverhu:
merged.. we can iterate if this is not enough.

arde171:
Sounds good.

Thanks,
Arup

oliverhu:
@arde171 i think you also need to bump version so that you can release it to Nexus

arde171:
Sure.

Thanks,
Arup

PR #483, opened 2020-11-11
bump up hadoop version to 2.10.0

arde171:
bump up hadoop version to 2.10.0 and updated _testGetContainerEnvForDocker_ test expected result as the docker runtime classes available in this version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,7 +92,7 @@ ext.deps = [
 
 allprojects {
   group = "com.linkedin.tony"
-  project.version = "0.3.35"
+  project.version = "0.4.0"
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -33,7 +33,7 @@ nexusStaging {
 }
 
 
-def hadoopVersion = "2.7.3"
+def hadoopVersion = "2.10.0"
```
goyalankit, line 36: Upgrade the minor version of tony as well. This doesn't seem backward compatible. 

0.3.35 -> 0.4.0

hungj: +1

goyalankit (approved):

PR #489, opened 2020-12-11
Add more logging in TonY client and related utils to see if TonY config can be overridden by other parallel jobs.

burgerkingeater:
We already have a PR(https://github.com/linkedin/TonY/pull/455) to address the issue where concurrent TonY jobs could override each other's local TonY config. However we are seeing  TonY job config is conflicting again. This PR adds more logging to help debugging this issue.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -221,6 +221,7 @@ public class TonyClient implements AutoCloseable {
     // Write user's overridden conf to an xml to be localized.
     try (OutputStream os = new FileOutputStream(tonyFinalConf)) {
       tonyConf.writeXml(os);
+      LOG.info("TonY conf " + tonyFinalConf + " is created locally.");
     } catch (IOException exception) {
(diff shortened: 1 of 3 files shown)


tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -221,6 +221,7 @@ public String processFinalTonyConf() throws IOException, ParseException {
     // Write user's overridden conf to an xml to be localized.
     try (OutputStream os = new FileOutputStream(tonyFinalConf)) {
       tonyConf.writeXml(os);
+      LOG.info("TonY conf " + tonyFinalConf + " is created locally.");
```
oliverhu, line 224: make all these LOG.DEBUG, also change all port reserver/release DEBUG

hungj (approved):

oliverhu (changes requested):
-1

burgerkingeater:
TonY config conflict is addressed by https://github.com/linkedin/TonY/pull/490/. So we don't need this debugging log any more.

PR #490, opened 2020-12-12
Uuidify the name of local directory generated by TonY client

burgerkingeater:
This PR is to address the directory naming conflict caused by different jobs of same job ID but within different sub-flows running in parallel.

E.g, the following 3 model training jobs which have the same name could override each other's TonY xml directory if without the fix. 
image


oliverhu (approved):

annyan09023:
Thanks!

PR #495, opened 2021-01-13
Proceed without failing if container resource paths do not exist

alipasha2019:
If container resource paths do not exist, we want to be able to still continue looking at the resource paths that were available.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -19,6 +19,7 @@ import com.linkedin.tony.util.HdfsUtils;
 import com.linkedin.tony.util.Utils;
 import com.linkedin.tony.util.VersionInfo;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
@@ -532,13 +533,16 @@ public class TonyClient implements AutoCloseable {
         continue;
       }
       for (String resource: resources) {
+        // If a hdfs classpath does not exist, we skip it rather than failing the job.
+        // This is because there are some cases where while constructing a ML flow, we have a hdfs classpath that might
+        // or might not exist depending on run time behavior. So we specify both paths in that case and disregard the
+        // non-existent once
         LocalizableResource lr;
-        // If a path does not exist, skip rather than failing only for hdfs classpaths.
         try {
           lr = new LocalizableResource(resource, fs);
-        } catch (IOException ex) {
-          LOG.info("Resource path does not exist for: " + resource);
+        } catch (FileNotFoundException ex) {
           if (hdfsClasspath.contains(resource)) {
+            LOG.warn("HDFS classpath does not exist for: " + resource);
             continue;
           } else {
             throw ex;
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -596,9 +596,11 @@ public class TonyClient implements AutoCloseable {
       }
       // Filter out original local file locations
       resources = tonyConf.getStrings(resourceKey);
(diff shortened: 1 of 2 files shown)


burgerkingeater:
@alipasha2019 thanks for the PR, can you add an unit test?

zuston:
Why not fast fail?

alipasha2019:
> Why not fast fail?

There are some cases where while constructing a flow we have a path that might or might not exist depending on run time behavior. So we specify both paths in those cases.

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -532,7 +532,18 @@ public void processTonyConfResources(Configuration tonyConf, FileSystem fs) thro
         continue;
       }
       for (String resource: resources) {
-        LocalizableResource lr = new LocalizableResource(resource, fs);
+        LocalizableResource lr;
+        // If a path does not exist, skip rather than failing only for hdfs classpaths.
+        try {
+          lr = new LocalizableResource(resource, fs);
+        } catch (IOException ex) {
```
burgerkingeater, line 539: Should this be FileNotFoundException?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -532,7 +532,18 @@ public void processTonyConfResources(Configuration tonyConf, FileSystem fs) thro
...
+          LOG.info("Resource path does not exist for: " + resource);
+          if (hdfsClasspath.contains(resource)) {
+            continue;
```
burgerkingeater, line 542: can you add logging here and remove the log message above?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -532,7 +532,18 @@ public void processTonyConfResources(Configuration tonyConf, FileSystem fs) thro
         continue;
       }
       for (String resource: resources) {
-        LocalizableResource lr = new LocalizableResource(resource, fs);
+        LocalizableResource lr;
+        // If a path does not exist, skip rather than failing only for hdfs classpaths.
```
burgerkingeater, line 536: Can you elaborate on the motivation of this change for hdfs classpath in the comment and the PR description?

burgerkingeater:
can you add a corresponding unit test in TestTonyE2E?

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -532,7 +532,18 @@ public void processTonyConfResources(Configuration tonyConf, FileSystem fs) thro
...
+        try {
+          lr = new LocalizableResource(resource, fs);
+        } catch (IOException ex) {
+          LOG.info("Resource path does not exist for: " + resource);
```
oliverhu, line 540: this should be `log.warn`

tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -393,11 +394,19 @@ public boolean init(String[] args) throws ParseException, IOException {
 
     // Set hdfsClassPath for all workers
     // Prepend hdfs:// if missing
-    hdfsClasspath = cliParser.getOptionValue("hdfs_classpath");
-    if (hdfsClasspath != null && !hdfsClasspath.startsWith(FileSystem.get(hdfsConf).getScheme())) {
-      hdfsClasspath = FileSystem.getDefaultUri(hdfsConf) + hdfsClasspath;
+    String allHdfsClasspathsString = cliParser.getOptionValue("hdfs_classpath");
+    if (allHdfsClasspathsString != null) {
+      String[] allHdfsClasspaths = allHdfsClasspathsString.split(",");
+      for (int i = 0; i < allHdfsClasspaths.length; i++) {
```
burgerkingeater, line 400: not fully understand why this is needed. Can you make it a method and add corresponding comment so that it's more readable?

burgerkingeater (approved):

Issue #500, opened 2021-02-04
TonY requests yarn config "yarn.io/gpu" on non-GPU clusters


bradmiro:
I am having an issue with TonY 0.4.3 where my clusters that are not configured with GPUs are failing because it can't find the yarn config "yarn.io/gpu":

```
Exception in thread "AMRM Callback Handler Thread" org.apache.hadoop.yarn.exceptions.YarnRuntimeException: org.apache.hadoop.yarn.exceptions.ResourceNotFoundException: The resource manager encountered a problem that should not occur under normal circumstances. Please report this error to the Hadoop community by opening a JIRA ticket at http://issues.apache.org/jira and including the following information:
* Resource type requested: yarn.io/gpu
* Resource object: <memory:4096, vCores:1>
* The stack trace for this exception: java.lang.Exception
	at org.apache.hadoop.yarn.exceptions.ResourceNotFoundException.<init>(ResourceNotFoundException.java:47)
	at org.apache.hadoop.yarn.api.records.Resource.getResourceInformation(Resource.java:268)
	at org.apache.hadoop.yarn.api.records.impl.pb.ResourcePBImpl.getResourceInformation(ResourcePBImpl.java:215)
	at com.linkedin.tony.ApplicationMaster$RMCallbackHandler.onContainersAllocated(ApplicationMaster.java:1059)
	at org.apache.hadoop.yarn.client.api.async.impl.AMRMClientAsyncImpl$CallbackHandlerThread.run(AMRMClientAsyncImpl.java:402)

After encountering this error, the resource manager is in an inconsistent state. It is safe for the resource manager to be restarted as the error encountered should be transitive. If high availability is enabled, failing over to a standby resource manager is also safe.
	at org.apache.hadoop.yarn.client.api.async.impl.AMRMClientAsyncImpl$CallbackHandlerThread.run(AMRMClientAsyncImpl.java:432)
Caused by: org.apache.hadoop.yarn.exceptions.ResourceNotFoundException: The resource manager encountered a problem that should not occur under normal circumstances. Please report this error to the Hadoop community by opening a JIRA ticket at http://issues.apache.org/jira and including the following information:
* Resource type requested: yarn.io/gpu
* Resource object: <memory:4096, vCores:1>
* The stack trace for this exception: java.lang.Exception
	at org.apache.hadoop.yarn.exceptions.ResourceNotFoundException.<init>(ResourceNotFoundException.java:47)
[...]

PR #502, opened 2021-02-05
Fix TonY requests yarn config "yarn.io/gpu" on non-GPU clusters #500

oliverhu:
Fix TonY requests yarn config "yarn.io/gpu" on non-GPU clusters #500. Add a try catch block when querying GPU resource.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
+++ b/tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
@@ -89,7 +89,6 @@ import org.apache.hadoop.yarn.security.client.ClientToAMTokenIdentifier;
 import org.apache.hadoop.yarn.security.client.ClientToAMTokenSecretManager;
 import org.apache.hadoop.yarn.util.AbstractLivelinessMonitor;
 import org.apache.hadoop.yarn.util.UTCClock;
-import org.apache.hadoop.yarn.util.resource.ResourceUtils;
 
 
 public class ApplicationMaster {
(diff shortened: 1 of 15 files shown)


tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -1053,10 +1054,16 @@ public void onContainersAllocated(List<Container> containers) {
       for (Container container : containers) {
         // Need to explicitly remove container requests from remoteRequestsTable in AMRMClient, otherwise
         // resources get double-requested (YARN-1902)
+        int numGPU = 0;
+        try {
+          numGPU = (int) container.getResource().getResourceInformation(Constants.GPU_URI).getValue();
+        } catch (ResourceNotFoundException e) {
+          LOG.debug("No GPU resource configured for the cluster. ", e);
```
ashahab, line 1061: Is there a better way to check for no GPU resource in cluster? Throwing and caching exception seems extreme. 
See Item 69 from Josh Bloch's Effective Java: [Use exceptions only for exceptional conditions](https://ahdak.github.io/blog/effective-java-part-9/)

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -410,7 +410,9 @@ public static void addEnvironmentForResource(LocalResource resource, FileSystem
   public static AMRMClient.ContainerRequest setupContainerRequestForRM(JobContainerRequest request) {
     Priority priority = Priority.newInstance(request.getPriority());
     Resource capability = Resource.newInstance((int) request.getMemory(), request.getVCores());
-    Utils.setCapabilityGPU(capability, request.getGPU());
+    if (request.getGPU() > 0) {
+      Utils.setCapabilityGPU(capability, request.getGPU());
```
ashahab, line 414: What happens with the prior behavior(Utils.setCapabilityGPU(capability, 0)?

plliao: The function just returns with no harm.

tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -1074,10 +1076,15 @@ public void onContainersAllocated(List<Container> containers) {
...
+        int numGPU = 0;
+        if (ResourceUtils.getResourceTypeIndex().containsKey(Constants.GPU_URI)) {
+          numGPU = (int) container.getResource().getResourceInformation(Constants.GPU_URI).getValue();
+        }
+
         amRMClient.removeContainerRequest(Utils.setupContainerRequestForRM(new JobContainerRequest(
             "", 1, container.getResource().getMemorySize(),
             container.getResource().getVirtualCores(),
-            (int) container.getResource().getResourceInformation(Constants.GPU_URI).getValue(),
+            numGPU,
```
ashahab, line 1087: Is there a unit-test that ensures that this codepath works? We need one if there isn't one. We need to test both cases(if(true) and if(false)).

ashahab (changes requested):

ashahab (approved):

ashahab:
Approved, merge when ready.

PR #503, opened 2021-02-07
Enable tensorboard port reuse option

zuston:
### why?
Encountered similiar problems with [PR](https://github.com/linkedin/TonY/pull/456), task executor(when is the role of chief) is still subject to tensorboard port conflict where port could be grabbed by other processes on the host before task process launches and after ports are closed.

### How?
In order to solve tensorboard port conflict fundamentally, I have submit the [PR](https://github.com/tensorflow/tensorboard/pull/4616) to support TB port reuse. Now we could enable it in TonY.


zuston:
Please check it @oliverhu @burgerkingeater ~

oliverhu (approved):

PR #510, opened 2021-02-22
Catch exception on getting HDFS delegation token, like standby namenode exception

zuston:
### Why
When a non-active namneode is set in the **tony.other.namenode** configuration, an exception will be thrown.
I think we should catch it.


zuston:
@oliverhu Please review it.

oliverhu (approved):

PR #517, opened 2021-03-24
[MINOR] Ignore updating task info connection error to prevent app failure

zuston:
### Why
As we know, TonY client will get AM status from RM in `monitorApplication`. But when encountering connection error on exec `updateTaskInfos` method, it will throw exception and client will finished and exit none-zero code. Actually, AM finished successfully.
This will cause our workflow system to misjudge the task status. We should only rely on RM's status recognition of the application.

PR #522, opened 2021-03-30
TonYClient ignore connection error to prevent app failure when sending AM stop signal

zuston:
### Why
Same as this [PR](https://github.com/linkedin/TonY/pull/517)

> As we know, TonY client will get AM status from RM in monitorApplication. But when encountering connection error on exec updateTaskInfos method, it will throw exception and client will finished and exit none-zero code. Actually, AM finished successfully.

> This will cause our workflow system to misjudge the task status. We should only rely on RM's status recognition of the application.

In this PR, client will also send stop signal to AM. But when AM has finished, this call will throw exception.
I think we should ignore its exception.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonyClient.java
@@ -1004,15 +1004,21 @@ public class TonyClient implements AutoCloseable {
       }
     }
 
-    if (amRpcClient != null) {
-      amRpcClient.finishApplication();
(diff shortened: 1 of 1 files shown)


zuston:
@oliverhu Please take a look.Thanks

oliverhu:
@zuston could you add more description what you are changing? before & after behavior?

zuston:
Done. Added in the description  @oliverhu

oliverhu (approved):

PR #527, opened 2021-04-15
Remove AMRM credentials on task executor

zuston:


oliverhu:
why do we need this ?

zuston:
I think AMRMToken should not exist in the task execcutor.

Refer to Flink code: https://github.com/apache/flink/blob/43883ccff94305a1d700233d8d9ac3278d18acad/flink-yarn/src/main/java/org/apache/flink/yarn/Utils.java#L562

oliverhu:
got it, makes sense.

oliverhu (approved):

PR #533, opened 2021-04-21
Introduce standalone runtime type

zuston:


oliverhu:
@zuston is this the replacement for single node training on AM?

zuston:
I think not. Replace single-one will change a lot in AM and notebook submmitter. I will do it in another PR and guarantee compatibility.

This PR only allow user to request standalone task executor to training and not on the AM. 

I think in the feature, single-one will be repalced by stand-alone type.

So it’s necessary both later and now.

oliverhu:
got it, please add an example & a test for this to demonstrate how to use this.

zuston:
@oliverhu Please review again. Thanks

oliverhu (approved):

PR #535, opened 2021-04-26
[Runtime] Make job fast fail when conf is illegal

zuston:


tony-core/src/main/java/com/linkedin/tony/runtime/HorovodRuntime.java
```diff
@@ -51,6 +51,9 @@
 public class HorovodRuntime extends MLGenericRuntime {
     private static final String DRIVER = "driver";
     private static final String WORKER = "worker";
+    private static final List<String> ILLEGAL_CONFIG_KEYS = Arrays.asList(
+            "tony.driver.instances"
+    );
```
oliverhu, line 56: This is very hard to understand... I think what you want to prevent is, you don't want users to have more than 1 driver?

zuston: Yes. Actually user shouldn't configure config about driver role.
Currently because we will overwrite the user configuration, the job will not throw an exception. But this behavior is actually contrary to user expectations, so I think fast fail is more reasonable.

oliverhu: Right, then it should be `tony.driver.*`? would regex matching makes more sense or over engineering..? like ["!tony.driver.*"]

zuston: Sounds great!

tony-core/src/main/java/com/linkedin/tony/runtime/HorovodRuntime.java
```diff
@@ -51,6 +52,9 @@
 public class HorovodRuntime extends MLGenericRuntime {
     private static final String DRIVER = "driver";
     private static final String WORKER = "worker";
+    private static final List<String> ILLEGAL_CONFIG_REGEXS = Arrays.asList(
+            "tony.driver\\.([a-z]+)"
+    );
```
oliverhu, line 57: This sounds generic, we can move to generic runtime

tony-core/src/main/java/com/linkedin/tony/runtime/HorovodRuntime.java
```diff
@@ -185,13 +189,31 @@ public boolean canStartTask(TonyConfigurationKeys.DistributedMode distributedMod
...
+                .flatMap(x -> x.stream()).collect(Collectors.toList());
+
+        if (CollectionUtils.isNotEmpty(illegalKeys)) {
+            log.error("Not allowed to configure illegal conf in Horovod Runtime. "
+                    + "Illegal keys: " + illegalKeys);
+            return false;
+        }
+        return true;
+    }
+
```
oliverhu, line 216: same here,  I think this functionality is generic.

oliverhu (approved):

PR #549, opened 2021-05-10
Prevent the running containers from stopping when AM crash

zuston:


zuston:
I will add some test cases on it laterly.

oliverhu (approved):

PR #551, opened 2021-05-15
Add interface comparable for TonyTask

daugraph:
In some cases, users may use `Collections` such as `ConcurrentSkipListSet` to cache `TonyTask`, which requires the implementation of the `Comparable` interface


oliverhu:
can you add a unit test? what is the use case by the way?

daugraph:
> can you add a unit test? 
thanks for the reviews, I've add a unit test. 
> what is the use case by the way?
we need to report `Metric` of every valid task to an external server periodically, so its necessary to `cache` valid task and traverse it periodically, in the meanwhile, AM will add task to the `cache` upon registerWorkerSpec processed. So we have to ensure `cache` thread safety, considering efficiency and thread safety, we chose `ConcurrentSkipListSet` to cache tasks, which requires the implementation of the interface `Comparable`

zuston:
LGTM. 
Thanks for your contribution

oliverhu (approved):

zuston (approved):

PR #554, opened 2021-05-18
Update protobuf-gradle-plugin version to 0.8.13

zuston:


oliverhu:
@zuston can you check why CI fails?

zuston:
@oliverhu  Sorry for the late reply. 
It looks like gradle version should be upgrade. But I’m not familiar with gradle, i will take some time to solve.

oliverhu:
no worries, take your time

PR #557, opened 2021-05-21
Fix logic error when judging registeredTask containing failed task

zuston:


tony-core/src/main/java/com/linkedin/tony/ApplicationMaster.java
```diff
@@ -1274,7 +1274,7 @@ private boolean startupFailed() {
     return completedFailedTasks.stream().anyMatch(t -> {
       String taskId = t.getTaskInfo().getName() + ":" + t.getTaskInfo().getIndex();
       LOG.debug("Failed task ID: " + taskId);
-      boolean existFailed = !session.getRegisteredTasks().contains(t);
```
zuston, line 1277: The type of object `t` is `TonyTask`. But `session.getRegisteredTasks()` returns a `Set<String>`, which contains `jobname:taskIndex`, like `worker:1`

oliverhu (approved):

PR #562, opened 2021-05-27
Make  AM and TaskExecutor runtime interface separate

zuston:
In current framework runtime abstraction, task executor's interface and am's interface are mixed together, which will cause programming troubles and abusing the class variables.
In this PR, let the task executor and am interface seperate.


oliverhu:
makes sense.. would be good if you can upload a workflow diagram after this change

zuston:
Rebase it @oliverhu

tony-core/src/main/java/com/linkedin/tony/Framework.java
```diff
@@ -0,0 +1,62 @@
...
+package com.linkedin.tony;
+
+import java.io.IOException;
+
+import org.apache.hadoop.conf.Configuration;
+
+import com.linkedin.tony.tensorflow.TonySession;
+import com.linkedin.tony.util.Utils;
+
+public class Framework {
```
oliverhu, line 25: add documentation for this class. `Framework` is a generic name, I don't know a better name either, so better add more documentation explaining

oliverhu (approved):

tony-core/src/main/java/com/linkedin/tony/Framework.java
zuston: Got it.

PR #563, opened 2021-05-27
Allow to specify sidecar tensorboard startup extra options

zuston:
Sometimes, we need to specify extra sidecar tb-startup params, like tensorboard `--reload-interval` or `--bind_all`


oliverhu:
@zuston it is getting a bit too much hard coding for this... I think we should just re-write tony-config.xml at submission time to inject a worker for Tensorboard, what do you think?

zuston:
Got it. Close this PR and i will submit new PR to refactor.

PR #565, opened 2021-06-24
TonY should throw exception when gpu resource is not found on cluster

UWFrankGu:
When user request to run a GPU job on a cluster without GPU resources, we should throw a runtime error.


annyan09023 (approved):
Thanks for the quick change

oliverhu (approved):

PR #566, opened 2021-06-30
Introduce pluggable runtime provider

zuston:


tony-core/src/main/java/com/linkedin/tony/FrameworkRuntimeProvider.java
```diff
@@ -15,50 +15,54 @@
...
-import com.linkedin.tony.runtime.TFRuntime;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.ServiceLoader;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 public class FrameworkRuntimeProvider {
```
oliverhu, line 26: @zuston it'd be good if you can write a dev guide (a readme file) if others want to support another framework in TonY.

oliverhu (commented):
looks like a mini dependency injection..

tony-core/src/main/java/com/linkedin/tony/FrameworkRuntimeProvider.java
zuston: Good idea

zuston:
> looks like a mini dependency injection..

Just refer to the implementation of the Hadoop file system that using the Java SPI.

tony-core/src/main/java/com/linkedin/tony/FrameworkRuntimeProvider.java
zuston: Developing doc will be attached later on another PR.

oliverhu (approved):

PR #571, opened 2021-07-07
Compatible with Hadoop 2.6.0-cdh5.11.0

zuston:


zuston:
I found the latest TonY code can't be compatible with Hadoop 2.6.0-cdh5.11.0. Fix it @oliverhu

oliverhu:
@zuston is there a use case for this? We can create a release branch for 2.6, but keep the master up with 2.10+

zuston:
We use the Hadoop 2.6.0 with TonY 0.3.20 version, it works well. And it's necessary to keep compatible with Hadoop 2.6.0, since importing some higher level api(like `ResourceUtils`), this leads to incompatibility.
This PR just to fix this. I think no need to create extra release branch.

tony-core/src/main/java/com/linkedin/tony/util/Utils.java
```diff
@@ -211,6 +210,15 @@ public static void setCapabilityGPU(Resource resource, int gpuCount) {
...
   }
 
+  /**
+   * Uses reflection to get memory size.
+   * @param resource  the request container resource
+   * @return the memory size
+   */
+  public static long getMemorySize(Resource resource) {
+    return HadoopCompatibleAdapter.getMemorySize(resource);
+  }
```
oliverhu, line 220: remove these functions from utils? just call the compatible adapter from code is good enough.

oliverhu (approved):

PR #592, opened 2021-08-31
[CI] Fix ci bug

zuston:


oliverhu:
Discussed offline to tag a image that has Python binary linked..

Thanks for catching this error!

zuston:
The command of `python` is missing in circleci image(circleci/openjdk:8-jdk), have to change from `python` to `python3`.

oliverhu (approved):

PR #595, opened 2021-09-03
Add regression test for issue #157

nevesnunes:
This test ensures that when `interrupt()` is called, the EventHandler
thread is not calling `dataFileWriter.close()`.

In order to have a reproducable test and avoid flakyness, we mock the
following methods to ensure the order and timming of the race condition:

- `dataFileWriter.close()` is locked on a latch until an `interrupt()`
is sent, if that interrupt happens inside this call, it will cause an
assertion fail;
- `interrupt()` unlocks that latch after executing the real
implementation;
- `eventHandlerThread` requires `dataFileWriter` to be set with the
mock, as well as `inProgressHistFile` to be set with a real path, to
pass some checks during `stop()`.

On my fork, I have branch `test-before-fix-157` to validate that this 
test fails on the commit before the fix, with the assertion fail on 
unexpected interrupt. Branch `test-157` contains the fix, allowing the 
test to pass.

### DIFF
--- a/tony-core/src/test/java/com/linkedin/tony/events/TestEventHandler.java
+++ b/tony-core/src/test/java/com/linkedin/tony/events/TestEventHandler.java
@@ -208,11 +208,11 @@ public class TestEventHandler {
     assertTrue(!emitterThread.isAlive());
     verify(writer).close();
 
-    fs.delete(jobDir, true);
+    Utils.cleanupHDFSPath(fs.getConf(), jobDir);
   }
 
   @AfterClass
   public void cleanUp() throws IOException {
-    fs.delete(jobDir, true);
+    Utils.cleanupHDFSPath(fs.getConf(), jobDir);
   }
 }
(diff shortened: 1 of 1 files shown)


tony-core/src/test/java/com/linkedin/tony/events/TestEventHandler.java
```diff
@@ -146,6 +154,63 @@ public void testDrainQueue() {
...
+    emitterThread.join();
+
+    if (gotUnexpectedInterrupt[0]) {
+      fail("Unexpected interrupt.");
+    }
+    assertTrue(!eventHandlerThread.isAlive());
+    assertTrue(!emitterThread.isAlive());
+    verify(writer).close();
+
+    fs.delete(jobDir, true);
```
zuston, line 211: Using `Utils.cleanupHDFSPath(fs.getConf(), jobDir);` maybe better ?

tony-core/src/test/java/com/linkedin/tony/events/TestEventHandler.java
```diff
@@ -146,6 +154,63 @@ public void testDrainQueue() {
...
+  }
+
   @AfterClass
   public void cleanUp() throws IOException {
     fs.delete(jobDir, true);
```
zuston, line 216: The same as above

zuston:
Good job @nevesnunes . LGTM, minor comments left. Thanks for your contribution.

zuston (approved):

zuston:
Merged. Thanks again! @nevesnunes

PR #598, opened 2021-09-24
TonyClient to create FileSystem from Path to support fully qualified HDFS path

helloworld1:
FileSystem.get() will only use default base URI for path. However when using fully qualified HDFS path, the base URI can be different from default.
This change part of the TonyClient to use path.getFileSystem() to get the path to handle fully qualified HDFS path.

See APA-50360 for details.

### DIFF
--- a/tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
+++ b/tony-cli/src/main/java/com/linkedin/tony/cli/ClusterSubmitter.java
@@ -57,7 +57,7 @@ public class ClusterSubmitter extends TonySubmitter {
     Path cachedLibPath = null;
     try (FileSystem fs = FileSystem.get(hdfsConf)) {
       cachedLibPath = new Path(fs.getHomeDirectory(), TONY_FOLDER + Path.SEPARATOR + UUID.randomUUID().toString());
-      Utils.uploadFileAndSetConfResources(cachedLibPath, new Path(jarLocation), TONY_JAR_NAME, client.getTonyConf(), fs,
+      Utils.uploadFileAndSetConfResources(cachedLibPath, new Path(jarLocation), TONY_JAR_NAME, client.getTonyConf(),
           LocalResourceType.FILE, TonyConfigurationKeys.getContainerResourcesKey());
       LOG.info("Copying " + jarLocation + " to: " + cachedLibPath);
       boolean sanityCheck = client.init(args);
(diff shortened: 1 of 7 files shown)


zuston:
@helloworld1 Nice catch! Could you help add some unit test to cover this case?

helloworld1:
> @helloworld1 Nice catch! Could you help add some unit test to cover this case?

Actually I don't know if this can be tested locally through unit test. FileSystem needs to be mocked in unit test, but FileSystem object needs to be dynamically created from Path object. 
If we do not mock FileSystem, it actually doesn't make differences when handling local files.

helloworld1:
The new commit refactored the Utils.java to not take FileSystem as parameter for some methods to avoid error of passing wrong FileSystem object.

zuston:
> > @helloworld1 Nice catch! Could you help add some unit test to cover this case?
> 
> @zuston
> Actually I don't know if this can be tested locally through unit test. FileSystem needs to be mocked in unit test, but FileSystem object needs to be dynamically created from Path object.
> If we do not mock FileSystem, it actually doesn't make differences when handling local files.

In `MiniCluster` class, it provides the dfscluster. So you could use it to mock multiple filesystems. like one path `file:///tmp/xxxx` and another path `hdfs://xxxx/xxxx`. Do you think right?

helloworld1:
@zuston Thanks for the suggestion. In order to really test the issue out, we need to set a foreign cluster.
In the local cluster "/yarn/resource" is equivalent to "hdfs://local-cluster/yarn/resource" in the test.
However, when accessing the file in non local cluster like "hdfs://other-cluster/yarn/resource", The FileSystem needs to be created from that Path to handle that path. Otherwise error will be thrown

23-09-2021 17:10:46 PDT modelTrainingJob INFO - Exception in thread "main" java.lang.IllegalArgumentException: Wrong FS: hdfs://ltx1-lassonn01.grid.linkedin.com:9000/jobs/messengerai/encrypted/encrypted_nudging_training/runs/1088562/private_intermediate_results/tf2_trainer/cacheLibrariesJob/tensorflow_tonyLib, expected: hdfs://ltx1-holdem

I don't know how feasible to use unit test to set up two clusters and use one cluster to access the other cluster to test this scenario E2E.

zuston (approved):

zuston:
@helloworld1 you are right. I will merge it and may we could add unit test on extra PR.

oliverhu:
Thanks for the contribution @helloworld1 !

PR #599, opened 2021-09-27
Prevent loss of root cause due to resetting the final state

zuston:


tony-core/src/main/java/com/linkedin/tony/TonySession.java
```diff
@@ -273,12 +273,12 @@ public void onTaskCompleted(String jobName, String jobIndex, int exitCode, Strin
     // short circuit and immediately stop training if a worker failed.
     if (exitCode != ContainerExitStatus.SUCCESS && exitCode != ContainerExitStatus.KILLED_BY_APPMASTER) {
       if (isChief(jobName, jobIndex) || shouldStopOnFailure(jobName) || isFailOnWorkerFailure()) {
-        trainingFinished = true;
         String diagnostic = "Exit status: " + exitCode;
         if (taskDiagnosticMsg != null) {
           diagnostic += ". Error msg: " + taskDiagnosticMsg;
         }
         setFinalStatus(FinalApplicationStatus.FAILED, diagnostic);
+        trainingFinished = true;
```
zuston, line 281: Setting the `trainingFinished` to true after `setFinalStatus` method will be better

oliverhu (approved):

PR #603, opened 2021-09-30
Introduce some job stop policies on TFRuntime

zuston:
### Why
On TFRuntime, I found two problems on tensorflow training on our production env.
1. Sometimes, when using tf estimator api, evaluator maybe wait the newest global step but chief has finished. So evaluator will hang.
2. Sometimes, worker will hang due to tf bugs, but chief has finished its task. So we need to kill workers which not finished.

> Besides do we need to make job failed when some worker or evaluator not finished? 

I think it can depends on user with config.

### How 
To fix the above problems, introduce two conf key
1. `tony.application.xxxx.run-alone.timeout` 
(like `tony.application.xxxx.run-alone.timeout=3600000`)

2. `tony.application.succeeded-on-run-alone-timeout.jobtypes` 
(like `tony.application.succeeded-on-run-alone-timeout.jobtypes=evaluator,worker`)


zuston:
@oliverhu Could you help check this ticket?

PR #608, opened 2021-10-12
Make untrackedTaskFailed volatile

zuston:


zuston:
Sorry for lacking description.
Bug introduced by PR #607 
the  param of `untrackedTaskFailed` will be used in two thread
1. main thread to check the app status in monitor method
2. heartbeat thread

To make the param visablity, it's necessary to make `untrackedTaskFailed` volatile

PR #626, opened 2021-12-15
Add support to mitigate log4j vulnerabilities from user imported log4…

UWFrankGu:
TonY Executor may pull additional user jars from HDFS. The jar may include log4j vulnerabilities (https://www.lunasec.io/docs/blog/log4j-zero-day/). This patch mitigates the vulnerability.

### DIFF
--- a/tony-core/src/main/java/com/linkedin/tony/TonySession.java
+++ b/tony-core/src/main/java/com/linkedin/tony/TonySession.java
@@ -443,7 +443,7 @@ public class TonySession {
       jvmArgs += " -Dlog4j2.formatMsgNoLookups=true";
       String tonyTaskExecutorJavaAgent = tonyConf.get(TonyConfigurationKeys.TASK_EXECUTOR_JAVA_AGENT, "");
       if (!tonyTaskExecutorJavaAgent.isEmpty()) {
-        jvmArgs += " -javaagent=" + tonyTaskExecutorJavaAgent;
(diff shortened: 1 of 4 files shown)


tony-core/src/main/java/com/linkedin/tony/TonySession.java
```diff
@@ -429,14 +429,24 @@ public TonySession build() {
...
       this.tonyConf = tonyConf;
       return this;
     }
+
+    // Appends default jvm arguments to task executor job
+    private String appendDefaultJVMArgs(String jvmArgs) {
+      jvmArgs += " -Dlog4j2.formatMsgNoLookups=true";
+      String tonyTaskExecutorJavaAgent = tonyConf.get(TonyConfigurationKeys.TASK_EXECUTOR_JAVA_AGENT, "");
+      if (!tonyTaskExecutorJavaAgent.isEmpty()) {
+        jvmArgs += " -javaagent=" + tonyTaskExecutorJavaAgent;
```
cjqiang-lnkd, line 446: should be ':', instead of '=', i.e. "-javaagent:$filepath"

cjqiang-lnkd (changes requested):

tony-core/src/main/java/com/linkedin/tony/TonySession.java
UWFrankGu: Good catch, thanks!

PR #628, opened 2021-12-17
Add jvm arguments to application master container to mitigate log4j v…

UWFrankGu:
This pr adds log4j mitigation for tony application master


tony-core/src/main/java/com/linkedin/tony/TonyClient.java
```diff
@@ -930,6 +930,12 @@ String buildCommand(long amMemory) {
         + ApplicationConstants.LOG_DIR_EXPANSION_VAR);
     // Add am jvm configuration
     String amJvm = tonyConf.get(TonyConfigurationKeys.TASK_AM_JVM_OPTS);
+    // Add default jvm configuration
+    arguments.add("-Dlog4j2.formatMsgNoLookups=true");
```
oliverhu, line 934: why not just use TonyConfigurationKeys.TASK_AM_JVM_OPTS ?

UWFrankGu: Good point, I've added the default options in a separate PR, closing this one.

PR #635, opened 2022-01-19
Make job fail fast when container starting failed

zuston:
### Why
Now when starting container failed in one nodemanager, tony AM can't find it and the job wont fail until reaching the registration timeout.

### How
We need to make job fail fast once container startup failed.


zuston:
@oliverhu Could you help review?

zuston:
By the way, i found that when requested resource is invalid, TonY wont request again. 
But I think TonY could request resource again when starting container failed, in some cases that some containers maybe marked as a lost node. 
But now the job will fail directly and has to wait all resources ready before starting.

oliverhu:
makes sense..

oliverhu (approved):

Issue #641, opened 2022-02-05
Configurable status when dependency times out


oliverhu:
Follow up from #621 . 
In some of the cases, we don't want to fail the whole job even though a dependency times out. For example, if `chief` succeeded and there is a `worker` hanging for 1 hour, users can configure the job to still pass.

```
tony.application.group.A = worker,chief
tony.application.dependency.evaluator.timeout.after.A = 3600
tony.application.dependency.evaluator.timeout.after.A.ignored = true
```
@zuston thoughts?

zuston:
Make sense

zuston:
Updated version on https://github.com/tony-framework/TonY/pull/643 @oliverhu

PR #642, opened 2022-02-05
Configurable status when dependency times out

zuston:
Introducing the new config of `tony.application.dependency.[X].timeout.after.[GROUP].ignored = true` to solve the #641.

When specifying the above conf, it will make this task type untracked.


oliverhu (approved):
Looks great! can you also bump the version in `build.gradle` and publish a new release @zuston ?

PR #648, opened 2022-02-25
Introduce the config of max-waiting-time when task executor register to AM

zuston:
### Why
Introduce this config make that the max-registry-time in task executor can be configurable.

Besides, to solve the network failure, this PR make the rpc of `registerCallbackInfo` can retry multiple times.


zuston:
@oliverhu

oliverhu (approved):

PR #652, opened 2022-03-08
Seperate the python subprocess log from task executor log

zuston:


tony-core/src/main/java/com/linkedin/tony/Constants.java
```diff
@@ -94,6 +94,9 @@
   public static final String AM_STDOUT_FILENAME = "amstdout.log";
   public static final String AM_STDERR_FILENAME = "amstderr.log";
 
+  public static final String TASK_EXECUTOR_PYTHON_STDERR_FILENAME = "python.err";
+  public static final String TASK_EXECUTOR_PYTHON_STDOUT_FILENAME = "python.out";
```
oliverhu, line 98: not necessarily Python, we can practically run anything with tony.. need a better name

zuston: subprocess.err? @oliverhu

oliverhu (approved):

PR #678, opened 2023-05-12
[FIX] Ignore non-existing role to avoid NPE under group-dependency-timeout mechanism

zuston:
If the dependency role is not in the tasks, it will throw NPE. 

The reason of ignoring the non-existing role rather than pre-checking is that sometimes we would set the dependencies' group in advance, specially for platform integration. In this case, it's better to ignore this role.


zuston:
PTAL @oliverhu

zuston:
The license conflicts, so I have to update it.

oliverhu (approved):
Thanks!

PR #680, opened 2023-09-01
improvement(client): throw the exception when submitting to Yarn

zuston:
Sometimes, if you specify the non-existing queue name, the client will exit but no any exception.


zuston:
cc @oliverhu

oliverhu (approved):

PR #681, opened 2023-09-04
Support placement constraint

zuston:


zuston:
PTAL @oliverhu . This has been tested in our internal hadoop3.2.2 cluster

oliverhu (approved):

oliverhu:
LGTM, @zuston is there a way add a test for this in one of the integration tests using miniyarn cluster?

zuston:
> LGTM, @zuston is there a way add a test for this in one of the integration tests using miniyarn cluster?

Yes. But this could be included in next PRs, which could be extended with multi hadoop3.x tests mechanism. 

How about merging this firstly?

oliverhu:
Sounds good

PR #682, opened 2023-10-11
feat: support placement constraint in app level

zuston:
In #681 , the placement constraint in executor level  has been supported. 
This PR is to introduce app level placement.

BTW, this has be validated in our Hadoop3.2.2 cluster.


zuston:
cc @oliverhu

oliverhu (approved):

